PR #2, opened 2018-03-15
RN npm module

asutula:
The idea here is to create a RN native module that exposes the methods from `Mobile.framework` to RN. This module should be install-able into a RN project using npm and, optionally, Cocoapods (as well as however you do it for Android).


sanderpick:
@asutula does this gets merged?

asutula:
Hey, no I'll close it for now. Need to rethink the best way to do this given the way the dependencies work.

PR #3, opened 2018-03-17
Feature/mobile photos api

sanderpick:
@asutula this should get us going on testing the UI, @andrewxhill I added some more bits and notes into the `PinPhoto` method, let me know where you guys landed with the thumbnail encoding stuff. Once you solve that, I can update this branch, or visa versa. 

- adds sql tables and indexes for photos
- adds interface for interacting with this table / db
- adds "photo set" top level hash to the table on pin
- adds get photos and get photo base 64 string to mobile node api, and tests


sanderpick:
One more comment, stubbed in a metadata object that gets added to the “photo set” directory

PR #4, opened 2018-03-17
Move thumb nailing out of -go so that we can deal with HEIF in native layer

andrewxhill:


repo/wallet/data.go
```diff
@@ -84,25 +85,27 @@ func PinPhoto(reader io.Reader, fname string, nd *core.IpfsNode, apiHost string)
...
+	t, err := ioutil.ReadAll(thumb)
+	if err != nil {
+		return nil, err
+	}
+
+	tb := bytes.NewReader(t)
+	tb.Seek(0, 0)
+
+	// add the thumbnail
+	addFileToDirectory(dirb, tb, "thumb.jpg", nd)
```
sanderpick: hey dude, nice! don't think you need to create a new reader here, just use thumb direct in addFileToDirectory

repo/wallet/data.go
```diff
@@ -4,14 +4,14 @@ import (
 	"io"
 	"time"
 	"bytes"
-	"image/jpeg"
+	_ "image/jpeg"
 	_ "image/png" // register other possible image types for decoding
 	_ "image/gif"
 	"encoding/json"
-	"image"
+	_ "image"
```
sanderpick: apparently it's bad practice to use the underscore method import unless you are just "loading codecs" - if you dont need these just delete em

asutula:
Ok, so I have the RN layer creating a JPEG thumb. The thumb is written to a path that can be passed into the Go layer. So the Go pin method should now take two path, one for the thumb and one for the full image.

asutula:
Ah cool, looks like that's the way the code is currently working. @andrewxhill you want to address @sanderpick's comments then we'll get this merged? Can't wait to try it on a real device!

repo/wallet/data.go
andrewxhill: okay, i guess we didn't need the orig reader copy either

repo/wallet/data.go
andrewxhill: done

andrewxhill:
okay, should be gtg

asutula:
Do you know how to generate the updated `Mobile.framework`?

PR #9, opened 2018-03-24
Adds Remote Adding/Pinning and associated new methods/APIs

carsonfarmer:


mobile/node.go
```diff
@@ -171,7 +170,29 @@ func (n *Node) Stop() error {
 	return nil
 }
 
-func (n *Node) PinPhoto(path string, thumb string) (string, error) {
+func (n *Node) AddPhotoRemote(path string, thumb string) ([]wallet.Hashed, error) {
```
asutula: Huh I don't see these changes reflected in the `Mobile.framework` header... Do we still need to re-generate `Mobile.framework`?

carsonfarmer: Hah, I haven't updated it yet :) Hence the WIP. But also, I'm going to change it 'back' a little bit. So there will be _minimal_ changes to the actual Mobile.framework header and associated wrapper stuff...

carsonfarmer:
that should now be updated with the new framework now @asutula. i haven't gone into textile-photos to update the wrapper stuff there yet though

asutula:
Cool, I can make updates to the app based on the new framework.

carsonfarmer:
saweet! so just let me know if that's enough information from the remote server aaron. right now, i'm mostly relying on errors to propagate up... but if you need header information for some reason, let me know and i can probably return that stuff as well

PR #10, opened 2018-03-24
WIP: basic p2p

andrewxhill:


sanderpick:
took a quick peek, wondering if you're able to get pretty far just using openbazaar's message sending... think i saw some of that in there so maybe you're taking their lead already, but i'm not sure what the goal is here either so maybe doesn't make sense after all, there's is for text-like messages

carsonfarmer:
yeh that is kinda what we need (arbitrary msgs between peers), so need to dig deeper into their stuff. for sure. right now, just testing mostly

PR #15, opened 2018-04-02
Split local and remote adds into separate methods on wallet

carsonfarmer:


asutula:
So once I call `LocalAddPhoto()`, i can immediately start using the returned hash to query for image data (before I call `RemoteAddPhoto`)?

carsonfarmer:
Yes, it _should_ just use the local version... _but_ the way I have it now, you'd still have to hang onto the thumb until after you call `RemoteAddPhoto()`, because it takes the same inputs.

carsonfarmer:
Another thing to watch for, is I'm not sure how go will handle these two functions running concurrently, because they are both using a reader over the image and thumb files. So we'll have to watch out for that.

carsonfarmer:
If its a problem, we can read in the whole image into a buffer and use that instead if we need to.

asutula:
I'm wondering how this relates to error handling... I can call `LocalAddPhoto()` and get a successful result. Then I call `RemoteAddPhoto()` and it fails. What am I supposed to do? Undo the local add?

asutula:
For questions like that, it would be good to have all this logic bundles inside a single asyc go function

carsonfarmer:
Right yeh, that's a good call. I'm just not yet sure how to do that (will look into async functions in go). Mostly, I'm not sure how to expose this within objective-c. I guess the other thing to try is to wrap the two calls in objective-c and handle that logic there?

carsonfarmer:
Here's a nice example of how we might do this in the go layer:
https://gobyexample.com/channels

carsonfarmer:
Yeh, so async in go is really nice. No need for callbacks really. However, I'm not entirely sure how to wrap this up so that it behaves how we want it in object-c or the js layer.

asutula:
I found a nice example of how to do callbacks:  https://medium.com/@matryer/tutorial-calling-go-code-from-swift-on-ios-and-vice-versa-with-gomobile-7925620c17a4

carsonfarmer:
Cool. I’ll take a look and a stab at implementing.

PR #20, opened 2018-04-06
refactor(mobile): refactor add photo, add encryption

sanderpick:
Refactors mobile node's `addPhoto` to return a multipart form object that can be used by the native OS, which is better suited to handle large file uploads in a mobile environment where connectivity, battery power, etc. are more unwieldy.

Other stuff: 

- remove local `gx` deps, use global gopath for these because they are content addressed, meaning diff versions cannot collide, pretty cool
- adds asymmetric curve25519 encryption to photo files

TODO:

- decrypt files on the way out when requested via mobile node


sanderpick:
okay @asutula, wip is removed:
- files are decrypted when fetched via the base64 endpoint
- sped up builds by _a lot_ by install `gx` deps on our docker image

PR #22, opened 2018-04-07
Adds mobile/node method for accessing the mnemonic phrase

andrewxhill:
…r's private key

### DIFF
--- a/mobile/node.go
+++ b/mobile/node.go
@@ -273,6 +273,16 @@ func (n *Node) GetPhotoBase64String(path string) (string, error) {
 	return bs64, nil
 }
 
+// provides the user's recovery phrase for their private key
+func (n *Node) GetRecoveryPhrase() (string, error) {
+	mnemonic, err := n.node.Datastore.Config().GetMnemonic()
+	if err != nil {
+		return "", err
+	}
+
+	return mnemonic, nil
+}
+
 func (n *Node) unmarshalPrivateKey() (libp2p.PrivKey, error) {
 	kb, err := n.node.Datastore.Config().GetIdentityKey()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


mobile/node.go
```diff
@@ -273,6 +273,16 @@ func (n *Node) GetPhotoBase64String(path string) (string, error) {
...
 }
 
+// provides the user's recovery phrase for their private key
+func (n *Node) GetRecoveryPhrase() (string, error) {
+	mnemonic, err := n.node.Datastore.Config().GetMnemonic()
+	if err != nil {
+		return "", err
+	}
+
+	return mnemonic, nil
```
sanderpick: could just `return n.node.Datastore.Config().GetMnemonic()` but nbd

mobile/node_test.go
```diff
@@ -69,6 +70,17 @@ func TestNode_GetPhotoBase64String(t *testing.T) {
...
+	mnemonic, err := textile.GetRecoveryPhrase()
+	if err != nil {
+		t.Errorf("failed to create new recoveryphrase: %s", err)
+	}
+	list := strings.Split(mnemonic, " ")
+	if len(list) != 24 {
+		t.Errorf("got bad mnemonic length: %s", len(list))
+	}
+}
+
```
sanderpick: sweet, could also add a test that tries to recover the priv key from this...
```
seed := bip39.NewSeed(mnemonic, "") // not using a pass here
identityKey, err := repo.IdentityKeyFromSeed(seed, nBitsForKeypair)
```
then would need to unmarshal that byte slice as a priv key and compare, or maybe compare bytes?. also, our `repo` package doesn't currently export `identityKeyFromSeed`... prob better to put this test in repo_test anyway.

PR #23, opened 2018-04-07
Adds working confirmation code embeded in qr code for testing with app

andrewxhill:


andrewxhill:
okay, should be gtg

sanderpick:
fixed on my desktop branch, it's a version mismatch in the astilectron packages... only the main repo is versioned and it's very old, so have to lock it to latest on master in `Gopkg.toml`

PR #24, opened 2018-04-08
Desktop node

sanderpick:
Bunch of refactoring to enable the creating a desktop specific node.


sanderpick:
@andrewxhill @asutula here's the basic architecture of the new nodes. I've split up how a node gets started into 3 methods:

- `Start`: same as before on mobile, this brings the node online
- `ConfigureDatastore`: pass in an existing mnemonic phrase or `""` _if this is your first run_. On desktop this should be called when we receive the mobile peer's mnemonic phrase. Each call to this method will overwrite your existing private asymmetric encryption key pair
- `StartServices`: Desktop only. This start auto garbage collection and the http gateway.

sanderpick:
@asutula once this is merged, ping me if you'd like so we can go over new mobile node instantiation, there's a couple method name changes, should be able to see in framework headers, but happy to chat

andrewxhill:
did you want this still in WIP? just your comment to aaron left me wondering if it was ready

sanderpick:
oh right nice

PR #55, opened 2018-04-26
feat(cmd): bug fixes, peers cmd, also join rooms on mobile

sanderpick:


sanderpick:
gonna merge this cause master is currently half baked

PR #57, opened 2018-04-27
feat(core): share photo feature

sanderpick:
- Share a photo to a different thread via cli and mobile framework
- Also fixes a bug where the full res photo was being corrupted by reading the exif data (first handful of bytes) and then encrypting the rest (without resetting the reader), that was a fun one


sanderpick:
The `SharePhoto` method returns a multipart request payload like `AddPhoto`, thought we might want that at some point

PR #62, opened 2018-05-01
feat($mobile): mobile user apis

andrewxhill:
adds api methods for signup and in from mobile

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -45,7 +45,7 @@ import (
 )
 
 const VERSION = "0.0.1"
-const apiURL = "https://api.textile.io/"
+const apiURL = "https://api.textile.io"
 
 var client = &http.Client{}
 
--- a/core/core.go
+++ b/core/core.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
@@ -8,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path"
 	"path/filepath"
@@ -20,6 +22,7 @@ import (
 	"github.com/tyler-smith/go-bip39"
 	"gopkg.in/natefinch/lumberjack.v2"
 
+	models "github.com/textileio/textile-go/central/models"
 	"github.com/textileio/textile-go/net"
 	trepo "github.com/textileio/textile-go/repo"
 	"github.com/textileio/textile-go/repo/db"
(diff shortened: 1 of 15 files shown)


mobile/mobile.go
```diff
@@ -144,3 +147,69 @@ func (w *Wrapper) PairDesktop(pkb64 string) (string, error) {
...
+		return false, errors.New(fmt.Sprintf("weak password - crackable in %s", match.CrackTimeDisplay))
+	}
+	return true, nil
+}
+
+func (w *Wrapper) SignUp(username string, password string, identity string, identityType string) (int, *models.Response, error) {
+	if identityType === "" {
+		identityType = "email_address"
+	}
+	reg := map[string]interface{}{
```
sanderpick: you can just import the actual `Registration` model from central/models

mobile/mobile.go
```diff
@@ -144,3 +147,69 @@ func (w *Wrapper) PairDesktop(pkb64 string) (string, error) {
...
+// TODO: doesn't use a cleaned version of the phone number, if pn is supplied
+func (w *Wrapper) CheckPassword(password string, identity string) (bool, error) {
+	match := zxcvbn.PasswordStrength(password, []string{identity})
+	if match.Score < 3 {
+		return false, errors.New(fmt.Sprintf("weak password - crackable in %s", match.CrackTimeDisplay))
+	}
+	return true, nil
+}
+
+func (w *Wrapper) SignUp(username string, password string, identity string, identityType string) (int, *models.Response, error) {
```
sanderpick: could make this `SignUpWithEmail` to make this simpler on the RN side, then add separate phone method if we ever do that

mobile/mobile.go
```diff
@@ -144,3 +147,69 @@ func (w *Wrapper) PairDesktop(pkb64 string) (string, error) {
 
 	return topic, nil
 }
+
+// TODO: doesn't use a cleaned version of the phone number, if pn is supplied
+func (w *Wrapper) CheckPassword(password string, identity string) (bool, error) {
```
sanderpick: I'd say just let the REST API do this, so we keep the framework size down

mobile/mobile.go
```diff
@@ -144,3 +147,69 @@ func (w *Wrapper) PairDesktop(pkb64 string) (string, error) {
...
+	match := zxcvbn.PasswordStrength(password, []string{identity})
+	if match.Score < 3 {
+		return false, errors.New(fmt.Sprintf("weak password - crackable in %s", match.CrackTimeDisplay))
+	}
+	return true, nil
+}
+
+func (w *Wrapper) SignUp(username string, password string, identity string, identityType string) (int, *models.Response, error) {
+	if identityType === "" {
```
sanderpick: triple equals?

sanderpick:
you'll want this PR in first too https://github.com/textileio/textile-go/pull/61 since it has the referral stuff

core/core.go
```diff
@@ -42,6 +45,9 @@ import (
 )
 
 const VERSION = "0.0.1"
+const apiURL = "https://api.textile.io/"
```
sanderpick: i think the extra slash is duplicate in its current usage below, other than that, hell yeah

PR #64, opened 2018-05-02
fix($desktop): returns supported type for native modules

andrewxhill:
was return complex types, now converting them to json


andrewxhill:
going to cowboy merge

andrewxhill:
while sander's at the gym 🙈

PR #65, opened 2018-05-02
fix($mobile): returns strings instead of data

andrewxhill:
was hopeful that i could make data types work in both bridges, but finally just returns full on strings instead of data


andrewxhill:
this version is already in my pending textile-mobile pr

andrewxhill:
this needs a merge so we can do a new build with your double room join fix

PR #69, opened 2018-05-03
Start and Stop methods for gateway moved to node methods

carsonfarmer:
This should allow us to gracefully stop the gateway server before shutting down

PR #70, opened 2018-05-03
feat(core): local signin, signout, and central jwt token storage

sanderpick:
- stores username and jwt tokens in `ConfigDB` via a `SignIn` and `SignOut` internal core methods
- exposes methods for get getting username and tokens to RN
- runs all tests together, core + mobile node doing signup + signin to local running central api


carsonfarmer:
Dang, didn't mean to approve, I meant to request one change:

Please update line 176 in `textile.go` to reflect the new `NewNode` call signature.
```
node, err := core.NewNode(dataDir, false, logging.DEBUG)
```

sanderpick:
haha, was more involved than I imagined!

sanderpick:
ok, updated that call signature, also exposed a `SignOut` method which just ditches the local username and tokens

textile.go
```diff
@@ -173,7 +173,8 @@ func main() {
 
 	// create a desktop textile node
 	// TODO: darwin should use App. Support dir, not home dir
-	node, err := core.NewNode(dataDir, false, logging.DEBUG)
+	// TODO: make api url configuratable via an option flag
+	node, err := core.NewNode(dataDir, "https://api.textile.io", false, logging.DEBUG)
```
carsonfarmer: sweet 👍

mobile/mobile_test.go
```diff
@@ -38,6 +45,50 @@ func TestWrapper_StartAgain(t *testing.T) {
...
+}
+
+func TestWrapper_GetUsername(t *testing.T) {
+	un, err := wrapper.GetUsername()
+	if err != nil {
+		t.Errorf("get username failed: %s", err)
+		return
+	}
+	if *un != cusername {
+		t.Errorf("got bad username: %s", un)
```
carsonfarmer: Looks like this doesn't need to be `Errorf`?

central/controllers/controllers_test.go
```diff
@@ -1,10 +0,0 @@
-package controllers_test
-
-import (
```
carsonfarmer: do you need to import test here?

sanderpick: nah, file is deleted... cleaning up the test flow right now so it can find this test package in ci

mobile/mobile_test.go
sanderpick: oh yeah, needs to be `*un`

central/controllers/controllers_test.go
carsonfarmer: Are you going to add that to this PR or a separate one?

sanderpick: this one, tests won't pass otherwise

sanderpick: will ping you when i update it

sanderpick:
@asutula @andrewxhill some changes to `Mobile.framework`:
- `NewNode` requires the central api url as a second param (`"https://api.textile.io"`) - needed to do that for testing the whole shabang
- signup/in methods only return `error`, not the response object
  - `SignUpWithEmail(username string, password string, email string, referral string) error`
  - `SignIn(username string, password string) error`
- you can get username via `GetUsername`, returns `(string, error)`
- you can get the access token via `GetAccessToken`, returns `(string, error)`
- you can see if we're logged in with `IsSignedIn`, returns `bool`
- you can sign out with `SignOut`, returns `error`

sanderpick:
okay @carsonfarmer - fixes the tests. In the future, if tests are failing no need to even start the review, I can see what's failing and will just be working on fixing those issues... will expect the same from you guys!

carsonfarmer:
Ok, so I really do think it'll go better to merge #69 _before_ merging this one, particularly given the changes to `Mobile.framework` to come as a result...

core/core.go
```diff
@@ -305,6 +308,55 @@ func (t *TextileNode) Stop() error {
...
+		log.Errorf("signin error: %s", err)
+		return err
+	}
+	if res.Error != nil {
+		log.Errorf("signin error from central: %s", err)
+		return errors.New(*res.Error)
+	}
+
+	// local signin
+	if err := t.Datastore.Config().SignIn(creds.Username, res.Session.AccessToken, res.Session.RefreshToken); err != nil {
```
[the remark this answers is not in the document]
sanderpick: ok, adding!

PR #74, opened 2018-05-03
fix(gateway): recover from panic when double register

sanderpick:
Pulls out the `http.HandleFunc` call into its own method that we can call and recover from. The fact that it's still considered registered, must mean that own node instance is still ok, and the handler function _should_ be reusable. Fingers crossed :)


sanderpick:
ok @andrewxhill or @carsonfarmer if you're around... this prevents the double register and that crash on RN refresh

andrewxhill:
going to get this in my branch with jwt mobile. done but was hoping to get this android grid in the same go

PR #90, opened 2018-05-07
HTTP Gateway with randomized basic auth

carsonfarmer:
Basically, a hash map is maintained in memory, and passwords are randomized after each call to `getPhotos`.

Things to look for:
* Should we update only on `getPhotos`, or also each time a resource is fetched via the Gateway?
* Is the UUID random enough?
* Should we put back in the random port number on (re)startup?
* Should/can we turn on https for IOS?

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -2,12 +2,15 @@ package core
 
 import (
 	"context"
+	"crypto/rand"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/big"
+	mrand "math/rand"
 	"net/http"
 	"os"
 	"path"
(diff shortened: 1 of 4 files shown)


core/core.go
```diff
@@ -854,16 +855,24 @@ func (t *TextileNode) GetPhotos(offsetId string, limit int, album string) *Photo
...
 		Hashes: make([]string, len(list)),
+		Paths:  make([]string, len(list)),
 	}
 	for i := range list {
-		res.Hashes[i] = list[i].Cid
+		password := ksuid.New().String()
+		hash := list[i].Cid
+		t.HashPasses[hash] = password
+		res.Hashes[i] = hash
+		res.Paths[i] = hash + ":" + password + "@localhost" + t.GatewayProxy.Addr + "/ipfs/" + hash
```
sanderpick: do we need to pass `hash` three times?

sanderpick: ok cool. if you need the pieces of the path client side, then i'd just pass the pieces... 
```
type PhotoListItem struct {
	Hash  string `json:"hash"`
	Token string `json:"token"`
	Proto string `json:"proto"`
	Host  string `json:"host"`
}

type PhotoList struct {
	Items []PhotoListItem `json:"items"`
}
```

the idea wiht passing the whole path, was to make RN side easier, not harder! :)

sanderpick: let's just move on though, easy change later...

sanderpick:
happy to merge this but @carsonfarmer do me a favor an open an issue to clean this up after release (pass the pieces in a new list item type)

PR #96, opened 2018-05-08
Rotating passport 

andrewxhill:


andrewxhill:
so we're going to need this guy before/with sander's #98 or the asyncimages wont work

sanderpick:
@andrewxhill is this one ready? like you said, happy to get it in before #98

andrewxhill:
yeah, it could prob use a test somewhere, but if you see something open a ticket and i'll get you a follow-up asap

core/core.go
```diff
@@ -1115,40 +1155,31 @@ func (t *TextileNode) startGateway() (<-chan error, error) {
 	// try to register our handler
 	t.registerGatewayHandler()
 
-	// Determine port
-	port, err := t.GatewayPort()
-	if err != nil {
-		return nil, err
-	}
-	portString := fmt.Sprintf(":%d", port)
+	// pick a random port
```
sanderpick: what's happening here? are we still doing random? just a heads up, if we _are_ planning on doing random, should do it like this: https://github.com/textileio/textile-go/blob/sander/caption/repo/config/init.go#L233, which is from my `sander/caption` branch and is replacing the freeport stuff, which never really made sense since it only searched for free port in first init. Also, was seeing some invalid port errors on desktop now and again, since freeport was somehow picking out-of-bounds ports. Based on some research we want ports in the 1024-49151 range. So, this approach plus `crypto/rand` would be nice here... `math/rand` is fine for swarm addresses.

core/core_test.go
```diff
@@ -121,13 +121,13 @@ func TestTextileNode_ConnectToRoomPeers(t *testing.T) {
...
 }
 
-func TestTextileNode_GatewayPort(t *testing.T) {
-	_, err := node.GatewayPort()
-	if err != nil {
-		t.Errorf("add photo failed: %s", err)
-		return
-	}
-}
+// func TestTextileNode_GatewayPort(t *testing.T) {
```
sanderpick: delete?

core/core.go
```diff
@@ -863,6 +868,16 @@ func (t *TextileNode) SharePhoto(hash string, album string) (*net.MultipartReque
 	return t.AddPhoto(ppath, tpath, album)
 }
 
+func (t *TextileNode) GetHashRequest(hash string) HashRequest {
```
sanderpick: will open ticket for a test on this guy

core/core.go
carsonfarmer: yeh cool, we had random in there (obviously), but i think @andrewxhill convinced me it didn't really buy us much if an attacker was truly motivated... but could just do a single random port selection on initial init (i.e., do it if its handy, but not for security reasons)

carsonfarmer: haha, also, the random port selection _is_ done like that @sanderpick, i just took out the math and used hard-coded ranges :) thought not the _right_ ranges according to your experiments

sanderpick: yeah man, same page. It is like that minus the difference I mentioned :)

carsonfarmer: cool. so since we're _not_ doing random for now, do you think we should revert to taking the default gateway port and adding 1000 like we had before? or just hard-coding it like this for now is ok?

PR #97, opened 2018-05-08
feat(core): add photo and share captions

### LINKED ISSUE #94: Add caption to metadata object
I don't think this needs to be indexed in photos table

sanderpick:
fixes #94

- add caption as a new file in the photo directory during an add or share
- caption is accesible via `/ipfs/Qm.../caption`
- caption is indexed in `photos` table


sanderpick:
this one is g2g. would be good to have in release so we don't need a db table migration for captions, but will require a tiny change to the bridges: `SharePhoto(hash string, thread string, caption string)` <- caption added to signature

sanderpick:
(the relay stuff and some other work is in here too, fyi)

carsonfarmer:
I'm updating this guys, just FYI.

PR #98, opened 2018-05-09
refactor(core): allow sqlite operations when node is not started

sanderpick:


sanderpick:
Glad we went this way... much nicer actually.

core/core.go
```diff
@@ -937,6 +980,17 @@ func (t *TextileNode) GetFile(path string, album *trepo.PhotoAlbum) ([]byte, err
 	return b, err
 }
 
+// GetFileBase64 returns data encoded as base64 under an ipfs path
```
carsonfarmer: where does this get used?

core/core.go
```diff
@@ -984,6 +1041,7 @@ func (t *TextileNode) LoadPhotoAndAlbum(hash string) (*trepo.PhotoSet, *trepo.Ph
 	return ph, album, nil
 }
 
+// UnmarshalPrivatePeerKey returns a PrivKey instance from the base64 encoded ipfs peer key
```
carsonfarmer: yay docs!

core/core.go
```diff
@@ -1341,3 +1405,16 @@ func (t *TextileNode) handleHash(hash string, aid string, api iface.CoreAPI, dat
 	// check last hash
 	return t.handleHash(last, aid, api, datac)
 }
+
+func (t *TextileNode) touchDB() error {
```
carsonfarmer: yeh nice dude, this is great

mobile/mobile.go
```diff
@@ -91,43 +98,43 @@ func (w *Wrapper) SignIn(username string, password string) error {
...
+// SignOut calls core SignOut
 func (w *Wrapper) SignOut() error {
 	return tcore.Node.SignOut()
 }
 
+// IsSignedIn calls core IsSignedIn
 func (w *Wrapper) IsSignedIn() bool {
-	_, err := tcore.Node.Datastore.Config().GetUsername()
-	return err == nil
+	si, _ := tcore.Node.IsSignedIn()
```
carsonfarmer: much cleaner

core/core.go
sanderpick: currently not used, almost deleted it but left around in case it makes sense to get meta this way

core/core.go
```diff
@@ -888,8 +929,10 @@ func (t *TextileNode) GetPhotos(offsetId string, limit int, album string) *Photo
 	return res
 }
 
-// pass in Qm../thumb, or Qm../photo for full image
-// album is looked up if not present
+// GetFile cats data from the ipfs node.
+// e.g., Qm../thumb, Qm../photo, Qm../meta, Qm../caption
+// Note: album is looked up if not present
+// TODO: shouldn't this be available offline for local (pinned content)?
```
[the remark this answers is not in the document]
sanderpick: yeah, for sure... we'd have to create an "offline" / "local" node though, which should be fast... future enhancement

andrewxhill:
carson, i'm going to jump you here

andrewxhill:
oop, nevermind conflicts now

sanderpick:
yep, gimme 10

carsonfarmer:
Jump me?

sanderpick:
Force pushed after rebase... in which I cleaned up the server port determination. It's now just takes the `Addresses.Gateway` value in the config, which is random, but static, and created during node creation.

PR #102, opened 2018-05-10
feat(central): more node config for server profile

sanderpick:
- Needed some more control over the IPFS config to properly run our node on a machine with public ip addresses (note: if you run the central server locally via `docker-compose`, it won't be auto-discoverable because MDNS is off)
- Relay cache now just keeps each peer's latest update

### DIFF
--- a/repo/config/init.go
+++ b/repo/config/init.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/op/go-logging"
 
+	"gx/ipfs/QmatUACvrFK3xYg1nd2iLAKfz7Yy5YB56tnzBYHpqiUuhn/go-ipfs/repo"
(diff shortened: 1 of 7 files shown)


repo/config/init.go
```diff
@@ -32,6 +33,26 @@ var textileBootstrapAddresses = []string{
...
+	"/ip4/192.0.0.170/ipcidr/32",
+	"/ip4/192.0.0.171/ipcidr/32",
+	"/ip4/192.0.2.0/ipcidr/24",
+	"/ip4/192.168.0.0/ipcidr/16",
+	"/ip4/198.18.0.0/ipcidr/15",
+	"/ip4/198.51.100.0/ipcidr/24",
+	"/ip4/203.0.113.0/ipcidr/24",
+	"/ip4/240.0.0.0/ipcidr/4",
+}
+
```
andrewxhill: pro

PR #103, opened 2018-05-10
Split relay and central services

sanderpick:
Also prepends relayed updates with `relay:` so that multiple relays can ignore each others' relays.


sanderpick:
Heads up on this, we'll want to update the frameworks before I deploy a new version of the relay, since the mobile / desktop nodes need to have the handling for parsing the update message prefix off the ipfs cid.

PR #110, opened 2018-05-12
feat(mobile): notify mobile layer of thread updates

sanderpick:
also fixes some goroutine in loop bugs

### DIFF
--- a/mobile/mobile.go
+++ b/mobile/mobile.go
@@ -18,41 +18,54 @@ import (
 
 var log = logging.MustGetLogger("mobile")
 
+// Messenger is used to inform the bridge layer of new data waiting to be queried
+type Messenger interface {
+	Notify(update *tcore.ThreadUpdate)
+}
+
 // Wrapper is the object exposed in the frameworks
 type Wrapper struct {
-	RepoPath string
+	RepoPath  string
+	messenger Messenger
+}
+
+// NodeConfig is used to configure the mobile node
+type NodeConfig struct {
+	RepoPath      string
+	CentralApiURL string
+	LogLevel      string
 }
 
 // NewNode is the mobile entry point for creating a node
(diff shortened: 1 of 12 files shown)


mobile/mobile.go
```diff
@@ -18,41 +18,54 @@ import (
...
 	var m Mobile
-	return m.NewNode(repoPath, centralApiURL, logLevel)
+	return m.NewNode(config, messenger)
 }
 
 // Mobile is the name of the framework (must match package name)
 type Mobile struct{}
 
 // Create a gomobile compatible wrapper around TextileNode
-func (m *Mobile) NewNode(repoPath string, centralApiURL string, logLevel string) (*Wrapper, error) {
```
asutula: So how is log level working now?

mobile/mobile.go
```diff
@@ -246,10 +257,29 @@ func (w *Wrapper) PairDesktop(pkb64 string) (string, error) {
...
+	datac := make(chan tcore.ThreadUpdate)
+	go tcore.Node.JoinRoom(id, datac)
+	go func() {
+		for {
+			select {
+			case update, ok := <-datac:
+				if !ok {
+					return
+				}
+				w.messenger.Notify(&update)
```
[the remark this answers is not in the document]
asutula: awesome

mobile/mobile.go
sanderpick: It's the same but you set `LogLevel` on the config object instead. There were getting to be many constructor params.

sanderpick:
Refactored this to take a generic `Event` struct, not `core.ThreadUpdate`, which means a) mobile frameworks don't need `core` package, and `Messenger` can be used for any event. The bridges just pass on event name and the un-serialized payload object (`NSDictionary` / `WritableMap`)

sanderpick:
okay guys, this is g2g

PR #119, opened 2018-05-17
refactor exif libs and exif location info scrubbing

### LINKED ISSUE #68: Strip location exif from photos in shared Threads
We could later include optional fields in the off-image metadata model that is shared.

carsonfarmer:
a wip for dealing with sensitive exif data. will fix #68 when done


carsonfarmer:
@sanderpick so are you suggesting we just take out gps from exif all the time for now? not just on share?

sanderpick:
yes

carsonfarmer:
-[x] tests added, all exif scrubbed, photo orientation corrected
-[x] tested on desktop/dev machine - passing
-[x] tested on ios - passing
-[ ] tested on android

repo/photos/photos.go
```diff
@@ -42,40 +46,30 @@ func Add(n *core.IpfsNode, pk libp2p.PubKey, p *os.File, t *os.File, lc string,
...
 	var tm time.Time
-	var lat, lon float64 = -1, -1
-	x, err := exif.Decode(p)
-	if err == nil {
+	x, ok := exif.Decode(p)
+	if ok == nil {
 		// time taken
 		tmTmp, err := x.DateTime()
-		if err == nil {
+		if err != nil {
```
sanderpick: why this?

repo/photos/util.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/disintegration/imaging"
+	"github.com/rwcarlsen/goexif/exif"
+)
+
+// ImagePathWithoutExif makes a copy of image (jpg,png or gif) and applies
+// all necessary operation to reverse its orientation to 1
+// The result is a image with corrected orientation and without
+// exif data.
+// ReadImage makes a copy of image (jpg,png or gif) and applies
```
sanderpick: copy pasta?

repo/photos/photos.go
```diff
@@ -101,8 +95,11 @@ func Add(n *core.IpfsNode, pk libp2p.PubKey, p *os.File, t *os.File, lc string,
 	dirb := uio.NewDirectory(n.DAG)
 
 	// add the image
-	p.Seek(0, 0)
```
sanderpick: i see, so you're re-reading the file below so you don't need to rewind this... do we need to read the file twice? looks like the file will be opened a total of 3 times, including what happens in the `ImagePathWithoutExif` method, do you think that makes sense?

repo/photos/util.go
carsonfarmer: yup, fixed

repo/photos/photos.go
carsonfarmer: good catch 👍

repo/photos/photos.go
carsonfarmer: yeh, this is probably an issue. i could probably pass the reader directly to the `ImagePathWithoutExif` and `Seek` within there to read the exif data.

carsonfarmer:
I think this is now gtg @sanderpick. The fact that we drop exif data from the original input files keeps this pr much simpler, and I've tried to reduce the amount of 'noisy' changes, so this is quite simplified.

PR #121, opened 2018-05-17
Joining thread post node creation

andrewxhill:


core/core.go
```diff
@@ -566,6 +556,29 @@ func (t *TextileNode) IsSignedIn() (bool, error) {
 	return err == nil, nil
 }
 
+func (t *TextileNode) JoinThread(mnemonic string, name string) error {
```
sanderpick: Could we call this CreateAlbumByName or something like that? "join" clashes with the method "JoinRoom" and everything is still called albums in the code

sanderpick: ah, nvm, CreateAlbum already takes mnemonic and name...

sanderpick: Seems like this "setting the public room" is a pretty beta-y / mobile concept, we could just move the logic that says "overwrite it if exists" to mobile package in a very explicit method: SetPublicThread / Album

sanderpick: just thinking that we'll soon want a real "create thead" method that should not overwrite (so users can create their own threads), what you think?

mobile/mobile.go
```diff
@@ -150,6 +150,27 @@ func (w *Wrapper) IsSignedIn() bool {
 	return si
 }
 
+// SignOut calls core SignOut
+func (w *Wrapper) UpdateThread(mnemonic string, name string) error {
```
andrewxhill: @sanderpick if i just push this in here, that isolates this change a lot. pretty simple.

sanderpick: okay sweet

repo/db/albums.go
```diff
@@ -25,7 +25,7 @@ func (c *AlbumDB) Put(album *repo.PhotoAlbum) error {
 	if err != nil {
 		return err
 	}
-	stm := `insert into albums(id, key, mnemonic, name) values(?,?,?,?)`
+	stm := `insert or replace into albums(id, key, mnemonic, name) values(?,?,?,?)`
```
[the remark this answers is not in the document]
sanderpick: this feels dangerous here, can we just delete by name first, and then insert? anyway, will take a look later

andrewxhill:
okay, i think this fella is ready

andrewxhill:
ah right... i'm not familiar enough with how the server-side, docker build, relay setup works. but i suppose we could use similar build types to put the relay on the right room? i could use an assist here i guess.

core/core.go
```diff
@@ -793,6 +783,11 @@ func (t *TextileNode) CreateAlbum(mnemonic string, name string) error {
 	return t.Datastore.Albums().Put(album)
 }
 
+// DeleteAlbum removes an album record by id
+func (t *TextileNode) DeleteAlbum(id string) error {
```
sanderpick: Wait, why do you want another function for this?

andrewxhill: ah duh... gone

PR #126, opened 2018-05-23
fix(mobile): touchDB is needed in early calls that might hit db

andrewxhill:
updateThread was being called before db existed sometimes

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -1047,6 +1047,10 @@ func (t *TextileNode) GetMetaData(hash string, album *trepo.PhotoAlbum) (*photos
 	return data, nil
 }
 
+func (t *TextileNode) TouchDB() error {
+	return t.touchDB()
+}
+
 // GetLastHash return the caption under a hash
 // TODO: shouldn't this be available offline for local (pinned content)?
 func (t *TextileNode) GetCaption(hash string, album *trepo.PhotoAlbum) (string, error) {
--- a/mobile/mobile.go
+++ b/mobile/mobile.go
@@ -155,23 +155,20 @@ func (w *Wrapper) UpdateThread(mnemonic string, name string) error {
 	if mnemonic == "" {
 		return errors.New("mnemonic must not be empty")
 	}
-
-	ba := tcore.Node.Datastore.Albums().GetAlbumByName(name)
-	if ba == nil {
-		log.Debugf("creating album: %s", name)
-		err := tcore.Node.CreateAlbum(mnemonic, name)
-		if err != nil {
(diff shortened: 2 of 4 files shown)


mobile/mobile.go
```diff
@@ -155,23 +155,20 @@ func (w *Wrapper) UpdateThread(mnemonic string, name string) error {
...
-		}
-	} else {
-		log.Debugf("removing album: %s", name)
-		err := tcore.Node.Datastore.Albums().DeleteAlbum(ba.Id)
-		if err != nil {
-			log.Errorf("error deleting album %s: %s", name, err)
-			return err
-		}
-		return tcore.Node.CreateAlbum(mnemonic, name)
+	if err := tcore.Node.TouchDB(); err != nil {
```
andrewxhill: this is the only thing required to fix that fore/back/fore error that was happening every other time. Needed to wrap touchDB to export it as TouchDB since we have this all in Mobile not Core

mobile/mobile.go
```diff
@@ -155,23 +155,20 @@ func (w *Wrapper) UpdateThread(mnemonic string, name string) error {
...
+		return err
+	}
+	log.Debugf("deleting album if exists: %s", name)
+	err := tcore.Node.Datastore.Albums().DeleteAlbumByName(name)
```
andrewxhill: Adding a DeleteByName method just let me reduce the DB calls in this function.

carsonfarmer:
Confirmed working on iOS and Android.

repo/db/albums.go
```diff
@@ -91,6 +91,13 @@ func (c *AlbumDB) DeleteAlbum(id string) error {
 	return err
 }
 
+func (c *AlbumDB) DeleteAlbumByName(name string) error {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	_, err := c.db.Exec("delete from albums where name=?", name)
+	return err
+}
+
```
sanderpick: sweet, i'd say add this method to the interface as well, but it's totally ripped up in my branch, so may as well skip it.

mobile/mobile.go
sanderpick: good call

core/core.go
```diff
@@ -1047,6 +1047,10 @@ func (t *TextileNode) GetMetaData(hash string, album *trepo.PhotoAlbum) (*photos
 	return data, nil
 }
 
+func (t *TextileNode) TouchDB() error {
+	return t.touchDB()
```
sanderpick: nice nice, you could just make the original method public too

PR #130, opened 2018-05-31
fix(mobile): removes timers on mobile, repl. ping w/ connect

sanderpick:
Started seeing a lot of ping failures this morning on mobile. If the node can't connect with a common relay node, it's just hosed. So,
- replace ping with connect since once you're for sure connected, everything works perfectly
- _always_ try to connect before publishing anything
- remove tickers on mobile, they just make shutdown too fragile and we don't get much out of them anyway since most people won't be sitting there with the app open forever
- on mobile, publish latest whenever photos are requested

### DIFF
--- a/cmd/swarm.go
+++ b/cmd/swarm.go
@@ -12,11 +12,8 @@ import (
 
 	"github.com/textileio/textile-go/core"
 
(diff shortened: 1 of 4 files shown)


andrewxhill:
looks good to me

PR #133, opened 2018-06-03
Try own dht fork with DoneChan

sanderpick:
The idea here is that we want to know when we're done bootstrapping the DHT. Also adds some logic that ensures we don't try and publish / connect to other peers unless we know at least one peer... otherwise we're just spinning wheels.

### DIFF
--- a/core/core.go
+++ b/core/core.go
@@ -1158,7 +1158,7 @@ func (t *TextileNode) Publish(topic string, payload []byte) error {
 		return err
 	}
 	for _, o := range out {
-		log.Info(o)
+		log.Debug(o)
 	}
 	return t.IpfsNode.Floodsub.Publish(topic, payload)
 }
@@ -1281,7 +1281,7 @@ func (t *TextileNode) RepublishLatestUpdate(album *trepo.PhotoAlbum) {
 func (t *TextileNode) registerGatewayHandler() {
 	defer func() {
 		if recover() != nil {
-			log.Error("gateway handler already registered")
+			log.Debug("gateway handler already registered")
 		}
 	}()
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
--- a/core/core.go
+++ b/core/core.go
@@ -30,20 +30,21 @@ import (
 	"github.com/textileio/textile-go/repo/photos"
 	"github.com/textileio/textile-go/util"
 
-	utilmain "gx/ipfs/QmatUACvrFK3xYg1nd2iLAKfz7Yy5YB56tnzBYHpqiUuhn/go-ipfs/cmd/ipfs/util"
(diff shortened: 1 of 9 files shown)


core/core.go
```diff
@@ -398,6 +402,7 @@ func (t *TextileNode) Start() error {
 	} else {
 		log.Info("desktop node is ready")
 	}
+	t.online = true
```
sanderpick: using this flag to determine "online-ness" vs. just whether or not the ipfs node is nil, since we need to set it as non-nil in start _before_ blocking. so, instead we can wait till the very end to mark the textile node as online

package.json
```diff
@@ -40,9 +40,9 @@
   },
   "gxDependencies": [
     {
-      "hash": "QmatUACvrFK3xYg1nd2iLAKfz7Yy5YB56tnzBYHpqiUuhn",
+      "hash": "QmPrKYPftocu7r4DhE3LorFgsJLGUkTPFLUqkS8SsuAXYn",
       "name": "go-ipfs",
-      "version": "0.4.14"
+      "version": "0.4.16-dev"
```
sanderpick: in order to try this out, i forked from go-ipfs master, which is currently a whole minor version ahead of our version. hopefully we get some general improvements as well. (the hash above is our own gx published hash, which I have pinned to the relay node :)

core/core.go
```diff
@@ -205,10 +209,7 @@ func NewNode(config NodeConfig) (*TextileNode, error) {
 	// determine the best routing
 	var routingOption core.RoutingOption
 	if config.IsMobile {
-		// TODO: Determine best value for this setting on mobile
-		// cfg.Swarm.DisableNatPortMap = true
-		// NOTE: trying normal routing for a bit
-		routingOption = core.DHTOption
+		routingOption = core.DHTClientOption
```
sanderpick: been testing with this now that I think I understand the difference... in this mode, the node doesn't respond to node search requests. on mobile, the node is hardly ever awake, so seems kinda pointless. we should get some battery savings

andrewxhill: nic

core/core.go
```diff
@@ -387,6 +388,9 @@ func (t *TextileNode) Start() error {
 		}
 	}()
 
+	// wait for dht to bootstrap
+	<-dht.DefaultBootstrapConfig.DoneChan
```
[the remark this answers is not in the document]
andrewxhill: do we block the mobile ui on startup until this is complete? i was wondering, why not just move on and then block future requests until DoneChan? as in, have an intermediate state for the startup, while node.state == bootstrapping, all non-ipfs requests are ready to go.

core/core.go
andrewxhill: ah, sweet. this might be what i was asking

core/core.go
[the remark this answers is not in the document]
sanderpick: yep, that's pretty much how it is... you can do anything non-ipfs related after just creating the node. `Start` is really start IPFS

PR #135, opened 2018-06-05
feat(core): start the node offline, use channel for when online

sanderpick:
More complicated, but way better experience. The basic idea is to first start the textile node (create and IPFS node) in offline mode, then kickoff a different goroutine that constructs a new node IPFS node in online mode... when it's ready, swap out the offline node.

@andrewxhill FYI seeing a bunch of wrong password errors in textile.log, but nothing seems to be visibly wrong looking at the UI. Not sure if this is related to this change or your progressive loading branch, since that's what I've been testing this with.


andrewxhill:
read through it, looks good

PR #142, opened 2018-06-14
Signed thread blocks

### LINKED ISSUE #139: Sign + verify block updates
At first these will go out over Floodsub, but eventually will be broadcasted via own DHT. The JWT Ed25519 signing method is already in place.

### LINKED ISSUE #140: Add back periodic relay connect
I took this out for simplicity during the re-arch, but definitely needed under the current Floodsub setup.

sanderpick:
fixes #140 
fixes #139


andrewxhill:
reviewing now

PR #149, opened 2018-06-26
feat(api): stubbing in referral counter

andrewxhill:
adding counter to referral codes to start issuing them to users

### DIFF
--- a/central/controllers/referrals.go
+++ b/central/controllers/referrals.go
@@ -24,24 +24,29 @@ func CreateReferral(c *gin.Context) {
 	count := 1
 	// how many times should we be able to use them
 	limit := 1
+	// simple reference to who requested the new code
+	requested_by := ""
 	params := c.Request.URL.Query()
-	if params["limit"] != nil {
-		tmp, err := strconv.ParseInt(params["limit"][0], 10, 64)
-		if err == nil {
-			limit = int(tmp)
-		}
-	}
 	if params["count"] != nil && len(params["count"]) > 0 {
 		tmp, err := strconv.ParseInt(params["count"][0], 10, 64)
 		if err == nil {
 			count = int(tmp)
 		}
 	}
+	if params["limit"] != nil {
+		tmp, err := strconv.ParseInt(params["limit"][0], 10, 64)
+		if err == nil {
+			limit = int(tmp)
+		}
+	}
+	if params["requested_by"] != nil {
+		requested_by = params["requested_by"][0]
+	}
 
 	// hodl 'em
 	refs := make([]string, count)
 	for i := range refs {
-		code, err := createReferral(limit)
+		code, err := createReferral(limit, requested_by)
 		if err != nil {
 			c.JSON(http.StatusInternalServerError, gin.H{"error": err.Error()})
 			return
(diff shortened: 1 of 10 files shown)


central/controllers/referrals.go
```diff
@@ -1,15 +1,16 @@
...
 	"math/rand"
 	"net/http"
 	"os"
 	"strconv"
 	"time"
+
+	"github.com/gin-gonic/gin"
+	"github.com/globalsign/mgo/bson"
+	"github.com/textileio/textile-go/central/dao"
+	"github.com/textileio/textile-go/central/models"
```
andrewxhill: not sure why the imports moved... 🤷‍♂️

central/controllers/referrals.go
```diff
@@ -21,7 +22,15 @@ func CreateReferral(c *gin.Context) {
 
 	// how many should we make?
 	count := 1
+	// how many times should we be able to use them
+	limit := 1
```
andrewxhill: takes a `limit` param in the api, defaults to 1 which will be incremented down to zero when the code is used

central/dao/dao.go
```diff
@@ -109,7 +110,7 @@ func (m *DAO) FindReferralByCode(code string) (models.Referral, error) {
 // List referrals
 func (m *DAO) ListUnusedReferrals() ([]models.Referral, error) {
 	var refs []models.Referral
-	err := db.C(referralCollection).Find(bson.M{"used": bson.M{"$eq": nil}}).All(&refs)
+	err := db.C(referralCollection).Find(bson.M{"remaining": bson.M{"$gt": 0}}).All(&refs)
```
andrewxhill: unused would now just be any with Remaining count greater than 0

central/models/user.go
```diff
@@ -40,9 +41,8 @@ type Identity struct {
...
 type Referral struct {
-	ID      bson.ObjectId  `bson:"_id" json:"id"`
-	Code    string         `bson:"code" json:"code"`
-	Created time.Time      `bson:"created" json:"created"`
-	Used    *time.Time     `bson:"used" json:"used"`
-	UserID  *bson.ObjectId `bson:"user_id" json:"user_id"`
+	ID        bson.ObjectId `bson:"_id" json:"id"`
+	Code      string        `bson:"code" json:"code"`
+	Created   time.Time     `bson:"created" json:"created"`
+	Remaining int           `bson:"remaining" json:"remaining"`
```
andrewxhill: no longer a Used or UserID param on the referrals since they will be used by multiple users

sanderpick:
looking good dude

central/controllers/referrals.go
sanderpick: for this, if you remove the whitespace in the import blocks, then run `npm run format` it should brings it back

andrewxhill:
okay, so was thinking of adding the referral creation to the user signup step, but i'm tempted instead to just leave it as an api and then throw it into react native to grab a new one the first time it's needed.

sound okay?

andrewxhill:
Okay, this is gtg

sanderpick:
hell yeah. since we don't have ci setup for this yet, i'll have to manually build the docker-compose stack, push that to docker hub, and update the swarm... can do in a bit

PR #154, opened 2018-07-02
Textile Service, Thread Invites, and Devices

### LINKED ISSUE #152: Needed node APIs for thread management
Starting down the path of implementing thread creation and management. Looks like we need some additional APIs:
- [x] Join existing thread (will the existing `addThread` just work for this? YES)
- [x] List all threads I participate in
- [x] Thread metadata (members, invite link, etc)
- [x] Leave thread

sanderpick:
Removes pubsub as the primary means of peer communication in favor of direct p2p.

also fixes #152


sanderpick:
- [x] make a single key pair
- [x] add devices table
- [x] when add new thread, create invite for each device
- [ ] add offline message handling

andrewxhill:
damn nice, just poked through this guy.

sanderpick:
getting there! can't wait to try the offline stuff

sanderpick:
Happy with this... will do the offline stuff in a new PR because would like to work on updating the bridges with this guy so @asutula can get going on new thread api stuff.

sanderpick:
(@asutula this includes mobile apis for listing, adding, removing threads and devices)

asutula:
Ok, tons of changes in here... Any files I should be looking at more closely? If so, just leave some comment in them?

Makefile
```diff
@@ -15,6 +15,12 @@ android_framework:
 	gomobile bind -target=android -o textilego.aar github.com/textileio/textile-go/mobile github.com/textileio/textile-go/net
 	cp -r textilego.aar ~/github/textileio/textile-mobile/android/textilego/
 
+P_TIMESTAMP = Mgoogle/protobuf/timestamp.proto=github.com/golang/protobuf/ptypes/timestamp
+P_ANY = Mgoogle/protobuf/any.proto=github.com/golang/protobuf/ptypes/any
+PKGMAP = $(P_TIMESTAMP),$(P_ANY)
+protos:
+	cd pb/protos && PATH=$(PATH):$(GOPATH)/bin protoc --go_out=$(PKGMAP):.. *.proto
```
sanderpick: now using protobuffs instead of JWTs over the wire... was already part of the OB net setup, so went with it

cmd/devices.go
```diff
@@ -0,0 +1,74 @@
+package cmd
```
sanderpick: adds a `devices` command

mobile/mobile.go
```diff
@@ -40,32 +38,47 @@ type Messenger interface {
 	Notify(event *Event)
 }
 
-// Wrapper is the object exposed in the frameworks
-type Wrapper struct {
```
sanderpick: this includes the mobile `Wrapper` removal

mobile/mobile.go
```diff
@@ -40,32 +38,47 @@ type Messenger interface {
...
-// NewNode is the mobile entry point for creating a node
-// NOTE: logLevel is one of: CRITICAL ERROR WARNING NOTICE INFO DEBUG
-func NewNode(config *NodeConfig, messenger Messenger) (*Wrapper, error) {
-	var m Mobile
-	return m.NewNode(config, messenger)
+// Mobile is the name of the framework (must match package name)
+type Mobile struct {
+	RepoPath  string
+	Mnemonic  string
+	Online    <-chan struct{} // not readable from bridges
```
sanderpick: used by tests that need to run when online... the bridge / rn layer get's notified via the `messenger`

mobile/mobile.go
```diff
@@ -40,32 +38,47 @@ type Messenger interface {
...
 
-// Blocks is a wrapper around a list of Blocks, which makes decoding json from a little cleaner
-// on the mobile side
+// Threads is a wrapper around a list of Threads
+type Threads struct {
+	Items []ThreadItem `json:"items"`
+}
+
+// DeviceItem is a simple meta data wrapper around a Device
+type DeviceItem struct {
```
sanderpick: similar to `Blocks`, a json-able object around a list of devices

mobile/mobile.go
```diff
@@ -40,32 +38,47 @@ type Messenger interface {
...
+	RepoPath  string
+	Mnemonic  string
+	Online    <-chan struct{} // not readable from bridges
+	messenger Messenger
 }
 
-// Mobile is the name of the framework (must match package name)
-type Mobile struct{}
+// ThreadItem is a simple meta data wrapper around a Thread
+type ThreadItem struct {
```
sanderpick: same here, similar to `Blocks`, a json-able object around a list of threads

mobile/mobile.go
```diff
@@ -158,52 +174,108 @@ func (w *Wrapper) SignIn(username string, password string) error {
...
 }
 
 // GetAccessToken calls core GetAccessToken
-func (w *Wrapper) GetAccessToken() (string, error) {
+func (m *Mobile) GetAccessToken() (string, error) {
 	return tcore.Node.Wallet.GetAccessToken()
 }
 
+// Threads lists all threads
+func (m *Mobile) Threads() (string, error) {
```
sanderpick: used to list all threads. @asutula this is the main file to pay attention to

net/service.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/textileio/textile-go/pb"
+	inet "gx/ipfs/QmXfkENeeBvh3zYA51MaSdGUdBjhQ99cP5WQe8zgr6wchG/go-libp2p-net"
+	"gx/ipfs/QmZoWKhxUmZ2seW4BzX6fJkNR8hh9PsGModr7q171yq2SS/go-libp2p-peer"
+)
+
+var (
+	DuplicateMessage = errors.New("duplicate message")
+)
+
+type NetworkService interface {
```
sanderpick: this is the highest level abstraction for the network service. send and handle messages from other peers

net/service/handlers.go
```diff
@@ -0,0 +1,438 @@
...
+		return nil
+	}
+}
+
+func (s *TextileService) handlePing(pid peer.ID, pmes *pb.Message, options interface{}) (*pb.Message, error) {
+	log.Debugf("received PING message from %s", pid.Pretty())
+	return pmes, nil
+}
+
+func (s *TextileService) handleThreadBlock(pid peer.ID, pmes *pb.Message, options interface{}) (*pb.Message, error) {
```
sanderpick: this is the important method in here... handles receiving a `block` from a peer

net/service/handlers.go
```diff
@@ -0,0 +1,438 @@
...
+		return nil, errors.New("TODO")
+	case pb.Block_LIKE:
+		return nil, errors.New("TODO")
+	}
+
+	// handle block
+	return nil, thrd.HandleBlock(signed.Id)
+}
+
+func (s *TextileService) handleFollow(pid peer.ID, pmes *pb.Message, options interface{}) (*pb.Message, error) {
```
sanderpick: I left follow, unfollow, chat in here cause we might want at some point

net/service/handlers.go
```diff
@@ -0,0 +1,438 @@
...
+	//if err != nil {
+	//	return nil, err
+	//}
+	//n := notifications.UnfollowNotification{notifications.NewID(), "unfollow", id.Pretty()}
+	//s.broadcast <- n
+	log.Debugf("received UNFOLLOW message from %s", id.Pretty())
+	return nil, nil
+}
+
+func (s *TextileService) handleOfflineAck(pid peer.ID, pmes *pb.Message, options interface{}) (*pb.Message, error) {
```
sanderpick: `handleOfflineAck` will be used when the offline handling is in place. in net speak, an "ACK" is an an acknowledgement message. not needed if the receiver is online in our case, since the sender will get a direct response. in offline, the ack is added and retrieved from the DHT... sorta like a text message (offline) vs. a phone call (online)

net/service/sender.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/textileio/textile-go/pb"
+	inet "gx/ipfs/QmXfkENeeBvh3zYA51MaSdGUdBjhQ99cP5WQe8zgr6wchG/go-libp2p-net"
+	ggio "gx/ipfs/QmZ4Qi3GaRbjcx28Sme5eMH7RQjGkt8wHxt2a65oLaeFEV/gogo-protobuf/io"
+	"gx/ipfs/QmZoWKhxUmZ2seW4BzX6fJkNR8hh9PsGModr7q171yq2SS/go-libp2p-peer"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+type sender struct {
```
sanderpick: `sender`, (no relation to rock climbing), wraps up handling around the i/o streams needed for sending and receiving between each peer

net/service/service.go
```diff
@@ -0,0 +1,198 @@
...
+	"io"
+	"sync"
+	"time"
+)
+
+var log = logging.MustGetLogger("service")
+
+var ProtocolTextile protocol.ID = "/textile/app/1.0.0"
+
+type TextileService struct {
```
sanderpick: implements the `NetworkService` interface

pb/protos/block.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+option go_package = "pb";
+
+import "google/protobuf/timestamp.proto";
+
+message Block {
```
sanderpick: `Block` type proto message, that can be signed and verified (see below)

pb/protos/message.proto
```diff
@@ -0,0 +1,72 @@
+syntax = "proto3";
+option go_package = "pb";
+
+import "google/protobuf/any.proto";
+import "google/protobuf/timestamp.proto";
+
+message Message {
```
sanderpick: message wrapper, payload can be any other type, like a `Block`

repo/config/init.go
```diff
@@ -226,34 +218,6 @@ func defaultDatastoreConfig() native.Datastore {
 	}
 }
 
-// identityConfig initializes a new identity.
-func identityConfig() (native.Identity, error) {
```
sanderpick: the old way in which the IPFS identity (keys) were generated, now moved to a higher level because we need to be able to control the input seed (mnemonic)

repo/datastore.go
```diff
@@ -49,11 +48,29 @@ type ThreadStore interface {
 	DeleteByName(name string) error
 }
 
+type DeviceStore interface {
```
sanderpick: devices are stored locally. each time a thread is added, invites are sent to each paired device.

repo/datastore.go
```diff
@@ -49,11 +48,29 @@ type ThreadStore interface {
...
+	Queryable
+	Add(device *Device) error
+	Get(id string) *Device
+	GetByName(name string) *Device
+	List(query string) []Device
+	Delete(id string) error
+	DeleteByName(name string) error
+}
+
+type PeerStore interface {
```
sanderpick: `PeerStore` tries to keep track of who is in a thread because they each need to be contacted directly and sent updates. If this becomes cumbersome for large threads, we could institute some rules like:
- only send to 8 random peers
- if you get an update that is new to you, relay it to 8 random peers
This is sort of a cheap DHT... similar to what gundb does

sanderpick: That said, I don't think we'll need to deal w/ that for a good while

repo/db/profile.go
```diff
@@ -15,32 +15,6 @@ func NewProfileStore(db *sql.DB, lock *sync.Mutex) repo.ProfileStore {
 	return &ProfileDB{db, lock}
 }
 
-func (c *ProfileDB) Init(id string, secret []byte) error {
```
sanderpick: no longer needed because the master key is just the IPFS private key, which we can get direct from the ipfs config

wallet/net.go
```diff
@@ -0,0 +1,284 @@
...
+	"golang.org/x/net/context"
+	"gx/ipfs/QmZoWKhxUmZ2seW4BzX6fJkNR8hh9PsGModr7q171yq2SS/go-libp2p-peer"
+	libp2pc "gx/ipfs/QmaPbCnUMBohSGo3KnxEa2bHqyJVVeEEcwtqJAYxerieBo/go-libp2p-crypto"
+	"gx/ipfs/QmcZfnkapfECQGcLZaf9B79NRg7cRa9EnZh4LSbkCzwNvY/go-cid"
+	"time"
+)
+
+//var OfflineMessageWaitGroup sync.WaitGroup
+
+func (w *Wallet) sendMessage(message *pb.Message, peerId string) error {
```
sanderpick: wrapper around service, so that the wallet can directly call things like `sendMessage`

wallet/thread/thread.go
```diff
@@ -77,16 +74,123 @@ func NewThread(model *repo.Thread, config *Config) (*Thread, error) {
...
+		send:       config.Send,
 	}, nil
 }
 
+func (t *Thread) Close() {
+	close(t.updates)
+}
+
+// AddInvite creates an invite block for the given recipient
+func (t *Thread) AddInvite(target libp2pc.PubKey) (*model.AddResult, error) {
```
sanderpick: `AddInvite` creates a block of type invite, and sends it to peers

wallet/thread/thread.go
```diff
@@ -572,30 +559,52 @@ func (t *Thread) indexBlock(id string) (*repo.Block, error) {
...
+		return nil, err
 	}
-	token, err := jwt.NewWithClaims(crypto.SigningMethodEd25519i, claims).SignedString(t.PrivKey)
+	pkb, err := t.ipfs().PrivateKey.GetPublic().Bytes()
 	if err != nil {
-		return "", err
+		return nil, err
 	}
-	return token, nil
+	signed := &pb.SignedBlock{
```
sanderpick: `SignedBlock` goes out over the wire to each peer

wallet/net.go
```diff
@@ -0,0 +1,284 @@
...
+	if err != nil {
+		//if err := w.SendOfflineMessage(p, k, &message); err != nil {
+		return err
+		//}
+	}
+	return nil
+}
+
+// Supply of a public key is optional, if nil is instead provided n.EncryptMessage does a lookup
+//func (w *Wallet) SendOfflineMessage(p peer.ID, k *libp2pc.PubKey, m *pb.Message) error {
```
sanderpick: (one more comment), this is the method we'll need for offline... see line 26 above, if there was an error reaching a peer, send offline instead. it would be nice to know if a peer is mobile, that way maybe you'd just send them offline always

desktop/main.go
```diff
@@ -1,89 +1,259 @@
...
+
+func getQRCode() (string, string, error) {
+	// get our own public key
+	pk, err := textile.Wallet.GetPubKeyString()
+	if err != nil {
+		return "", "", err
+	}
+
+	// create a qr code
+	url := fmt.Sprintf("https://www.textile.io/clients?key=%s", pk)
```
andrewxhill: app requires a confirmation code e.g. 3425. the thought behind that was to show it on the desktop to ensure your phone correctly paired to the right device before sending keys out

andrewxhill: it'd just be code=3425

sanderpick: oh right, forgot i removed that. i think the address serves the same purpose and is more informative / secure... so when the phone goes to confirm the pairing it can just say, "about to pair with this address xxx" (the pub key below the qr code), and the user can double check if they want.

andrewxhill: perfect

sanderpick: we'll also need to let the mobile side name the device, which would presumably happen before the confirmation (?)

PR #158, opened 2018-07-06
Offline message handling

### LINKED ISSUE #151: getBlockData throws exception when data isn't found locally
I sometimes see the error coming from the go layer: `block not found locally`.

I think this error happens when they data you're querying for doesn't exist for a valid reason (like the user didn't comment on their photo). Need to do some more testing to totally understand the behavior, but let's decide what we think the correct behavior is... Return `null`? Something else?

### LINKED ISSUE #153: Add thread blocks for user joining/leaving thread
This way, state about who's currently participating in the thread can be shared among all members of the thread.

### LINKED ISSUE #155: CLI needs a `daemon` mode
Since the cli is interactive, it doesn't seem like like being started with `systemctl` or `&` as a background service. The only way i could get it to work for the public facing decrypting gateway was using `screen`.

### LINKED ISSUE #161: Thread related mobile APIs are thread name based
This makes the assumption that thread names are unique. I don't think we can assume that once we launch shared threads. These APIs should be thread id based.

### LINKED ISSUE #162: Threads.Items is null when no threads exist
Would be better if it were an empty array

### LINKED ISSUE #166: removeThread is by Name in Go
You can see it in the bridge 

https://github.com/textileio/textile-mobile/blob/master/ios/NativeModules/TextileNode.m#L315

and saga

https://github.com/textileio/textile-mobile/blob/master/App/Sagas/TextileSagas.ts#L360

but in go

https://github.com/textileio/textile-go/blob/master/mobile/mobile.go#L257

### LINKED ISSUE #167: PhotoBlocks.Items is null if no blocks returned
It should return an empty array

### LINKED ISSUE #185: Add image data url prefix to photo block data on mobile
... on `/thumb` and `/photo` links. This is currently handled by the the native bridges. Will be nicer to do here because we can more easily detect (guess) the file type. Saves having to pass file type to the `TextileImage` native component.

sanderpick:
Offline message handling for `TextileService`, based on OB. The basic idea is to store messages in the DHT if the recipient is offline. A `retriever` runs periodically, looking for new messages.

fixes #166
fixes #167 
fixes #161
fixes #162  
fixes #151 
fixes #153 
fixes #155 
fixes #185


sanderpick:
Got the pieces in place here, now to get it working...

sanderpick:
Got this working on the plane - couple bits to iron out though.

andrewxhill:
For the Thread invites, I went with a few more parameters including in the invite than we have talked about previously (https://github.com/textileio/textile-mobile/pull/178/files#diff-422240a0aedd992418cfb42e4e6cab18R14). Specifically, the device will now want the key and _also_ the name of the invitee as `from` and the provided name for the thread as `name`. This allows us to be a bit more detailed in the invite confirmation step *and* will allow us to provide a name immediately on joining.

sanderpick:
Ok, cool. Just to be clear, you are talking about external invites, correct? As in not pairing (both use the links). In this branch, the framework generates an external invite link that looks like:

`https://www.textile.io/clients/#invite=QmeHJZQYiRdHETequdgaPNs8BQd1wJwZCXvMWksXiFrTEH&key=17R74Tgcw9hC41x9fvaZQPtiRSA17R74QGR5hTRivzNj&name=boom`

(name is included, which is a suggested thread name... if you already have that name locally, it bumps to name_1, eg, `cats` -> `cats_1`, up to 100 tries, then will error out)

based on chats, was planning to:
- include an external invite host+path in the node config, swap that out
- replace the word `invite` with `id` since it looked like you wanted the word "invite" in the path
- my thinking here was you could add more params like referral/from at the application level, they aren't important at the go layer (will be ignored)

andrewxhill:
okay, that sounds good + makes sense

cmd/threads.go
```diff
@@ -160,25 +182,44 @@ func RemoveThread(c *ishell.Context) {
 	}
 	name := c.Args[0]
 
-	err := core.Node.Wallet.RemoveThread(name)
+	addr, err := core.Node.Wallet.RemoveThread(name)
```
andrewxhill: maybe just a variable naming fix here ^

wallet/thread/helpers.go
```diff
@@ -0,0 +1,64 @@
...
+func (t *Thread) GetBlockDataBase64(path string, block *repo.Block) (string, error) {
+	data, err := t.GetBlockData(path, block)
+	if err != nil {
+		return "error", err
+	}
+	return libp2pc.ConfigEncodeKey(data), nil
+}
+
+// GetPhotoMetaData returns photo metadata under an id
+func (t *Thread) GetPhotoMetaData(id string, block *repo.Block) (*model.PhotoMetadata, error) {
```
andrewxhill: Should this one be exposed to Mobile or is there a diff method to grab the meta file I should use?

cmd/threads.go
sanderpick: don't follow?

sanderpick: good catch

wallet/thread/helpers.go
sanderpick: you can use the generic `GetPhotoData` method... but perhaps we want to make it a bit more explicit for all things in mobile since from a client perspective it has no idea how the data is structures, so req'ing thing like `/photo` are a bit weird.

sanderpick: We could do:
`GetPhotoData`
`GetThumbData`
`GetMetaData`

sanderpick: That would actually help with the gif thing too.

andrewxhill: relates to this, https://github.com/textileio/textile-go/issues/170

sanderpick: ok, done. image methods will now return the correct data url prefix as well. `TextileImage` will have to be tweaked (delete the prefix: https://github.com/textileio/textile-mobile/blob/ea6addf98ef3bfa3daeeb135d1bf34626f5d5218/ios/NativeViews/TextileImageView.m#L65)

PR #165, opened 2018-07-15
Updates pairing url to use textile.photos

andrewxhill:
will put up a landing page to catch instances where the deep link ends up being rendered in a browser (similar to https://www.textile.photos/invites/new?key=1&from=m&referral=code) to help users troubleshoot


andrewxhill:
actually realized we do still need this

sanderpick:
cool, updating branch

PR #187, opened 2018-07-19
fix(mobile get photos): caption encrypted w/thead key, not photo key

sanderpick:
also adds version to metadata


sanderpick:
Looks like CircelCI is having an outage...

PR #188, opened 2018-07-19
Remote pinning service for thread blocks

sanderpick:
Pinning service that runs:
- on mobile, on start and online
- on desktop / server, on start and online, and every 10 min
- whenever a new update is made. so, in the ideal case the queue is at most length 1, then drops to 0 almost immediately 

Technically we don't need to be online (in libp2p host speak), but using that as an indication that we're likely able to open a socket.

@asutula @andrewxhill  Somewhat unrelated, we'll need some logic around when to call `RefreshMessages` (*new mobile method*) from the mobile app. This method runs internal service jobs that fetch new message pointers from the DHT, as well as re-publishing self-authored pointers. I could see calling it whenever a thread is "pulled down" or some refresh button is tapped (like in IG). Leaving that up to the client in the mobile case.


sanderpick:
Okay guys, take a look at this one... seems to be working nicely.

andrewxhill:
Sweet. What are you thinking the basic logic should be? For now maybe just each app foreground or something?

andrewxhill:
i'll merge and add a bridge for that new method

asutula:
I guess I need to hear more about `RefreshMessages`... is this the thing that will give the app "notifications" of different types?

asutula:
Or i guess that is what will fetch all new data? (images, invites, joins, comments, etc). If so, is that the only way to get new content? The mobile app has to poll? There's no push mechanism while the app is foregrounded?

sanderpick:
No, this is for messages delivered while you're offline (part of the offline service handling). They are entered as pointers directly into the DHT... there's no why to "push" those because there's no server to push from... you have to go out and retrieve them. So that retrieval is the job that needs to run. If it finds messages, they will be pushed over the event bus to the app. A message could be a new photo, invite, join, leave, etc... any thread block update.

sanderpick:
Sorta like IG or FB type apps, it will run whenever the node starts (see PR description for when it runs), then you can control whenever else you want it to run, like when somebody taps a refresh button.

sanderpick:
Oops, PR desc only mentioned when the pinning service runs... `RefreshMessages` after a start, and after the p2p host is online. From then on, you're in control.

sanderpick:
(which is a foreground)

PR #193, opened 2018-07-21
feat(service): use block as is for offline messages

sanderpick:
Previously, we were building a unique payload for each offline recipient, then asking the network to please store that (neighbors closest to the resultant hash key)... we're already pinning block updates remotely - so if we just use the block _as is_ for the message payload, no need to bother with re-storing the message somewhere.

Also,
- fixes a miriad of thread join bugs (back prop was totally broken, and joining an external didn't even try to back prop)
- fixes desktop pairing

MOBILE: tiny change - a thread update event's `index` field was re-named to `block` (shouldn't require a bridge update)


andrewxhill:
@asutula, I just breezed passed this, but in case you missed it and it has some effect. the comment on this PR 

"MOBILE: tiny change - a thread update event's index field was re-named to block (shouldn't require a bridge update)"

asutula:
Think we're fine, we only use the `thread_id` field.

PR #200, opened 2018-07-25
feat(profile): publish profile updates to ipns, add cmds

### LINKED ISSUE #168: Use IPNS for profile
- username
- avatar
- other info?

sanderpick:
- Adds methods for interacting with IPNS
- A simple profile is (re)published under peer id whenever login, logout happens, as well as on start up
- New gateway methods for getting at IPNS links

TODO: 
- Pinner needs to handle the whole profile directory
- Mobile methods for getting other peers' profiles, since the gateway isn't running

#fixes #168


sanderpick:
screenshot 2018-07-25 14 56 25

sanderpick:
screenshot 2018-07-25 15 00 59

mobile/mobile.go
```diff
@@ -493,6 +493,25 @@ func (m *Mobile) GetPhotoMetadata(id string) (string, error) {
 	return toJSON(meta)
 }
 
+// ResolveProfileInfo take a peer id and profile key name, like "username"
+func (m *Mobile) ResolveProfileInfo(peerId string, key string) (string, error) {
```
sanderpick: ^ here's the method for getting profile keys like username, and avatar_id under a peer's id

wallet/model/model.go
```diff
@@ -5,14 +5,15 @@ import "time"
...
-	AvatarId string `json:"avatar_id"`
+	AvatarId string `json:"avatar_id,omitempty"`
 }
 
 const ThumbnailWidth = 300
 
 type Metadata struct {
 	Version  string    `json:"version"`
-	Username string    `json:"username,omitempty"`
+	PeerId   string    `json:"peer_id"`
```
sanderpick: Adding peer id here because it's needed in order to resolve a peer's profile. I think we need to rethink the metadata structure here, ideally it would be a traversable DAG, but then we'd have to encrypt each link separately, and the link names would be plaintext... trade offs.

sanderpick:
(still need to fix the profile pin step... walk the dir and pin each...)

core/cafe/archive.go
```diff
@@ -2,30 +2,45 @@ package client
...
+	"bytes"
 	"compress/gzip"
 	"fmt"
+	"io"
 	"os"
-	"path/filepath"
 )
 
 type Archive struct {
-	Id   string `json:"id"`
```
sanderpick: @asutula `id` is now removed here

core/cafe/archive.go
```diff
@@ -44,12 +59,22 @@ func (a *Archive) AddFile(blob []byte, fname string) error {
 	return nil
 }
 
+func (a *Archive) VirtualReader() io.Reader {
```
sanderpick: `Archive`s can now be "virtual", which just means we're writing direct to a buffer (not a file)... useful for handling pin requests that are actually a directory structure (DAG). In that case, a virtual archive is created from the DAG links, which dumped to the remote pin API.

sanderpick: (this happens only when the pinner runs and reaches an entry that is a DAG)

util/ipfs.go
```diff
@@ -27,8 +30,53 @@ var log = logging.MustGetLogger("util")
...
+	reader, err := GetReaderAtPath(ipfs, path)
+	if err != nil {
+		return nil, err
+	}
+	return ioutil.ReadAll(reader)
+}
+
+// GetArchiveAtPath builds an archive from directory links under an ipfs path
+// NOTE: currently will bork if dir path contains other dirs (depth > 1)
+func GetArchiveAtPath(ipfs *core.IpfsNode, path string) (io.Reader, error) {
```
sanderpick: here's the method for building that archive

sanderpick:
Ok guys, damn that was rough... brain must be fried. Please take a look / merge this rig. We'll need a bridge update for the mobile method: `ResolveProfileInfo`

asutula:
Cool man, looks good from a high level. The only questions I have (which may not be relevant, just first reactions):

1. Since you have to know the profile keys to query each one of them, how does the client learn the keys?
2. Having to fetch single profile keys seems odd vs getting back a profile object
3. Image metadata seems like an odd place to get a 'global' thing like peer id. Like if we need to query for profile data, you have to go find some image metadata first, and get the peer id out of it, then make your profile query?

Like I said, maybe not actually relevant, maybe this is the way it should work with ipns etc

sanderpick:
Sweet, thanks for the look! 

> 
    1. Since you have to know the profile keys to query each one of them, how does the client learn the keys?
Good question, I'd say let's revisit when we move to protobufs on the interface. Feels similar to any other model we share... e.g, client needs to know the structure of meta data.
> 
    2. Having to fetch single profile keys seems odd vs getting back a profile object
Could definitely expose the full object if that makes more sense. I guess my thinking was more desktop based... where you want to just use the gateway, so having the key-based links is nice. Can do this in a follow up, will make issue.
>
    3. Image metadata seems like an odd place to get a 'global' thing like peer id. Like if we need to query for profile data, you have to go find some image metadata first, and get the peer id out of it, then make your profile query?
You can always get your own peer's id, username, etc. directly... it's just that the only time you'd encounter another peer is if they added a photo. It's just the same as an IG post likely has the user's ID on it. So, once you have their peer id, you can then find profile info, rather than putting all the info (ever expanding) right on the photo / comment / etc.

PR #212, opened 2018-07-30
feat(threads): adds ignore blocks

### LINKED ISSUE #172: Add Remove block type to Threads
The ability to remove a photo will need the ability to request its removal from a thread. We were thinking this would just be a remove block type.

related https://github.com/textileio/textile-go/issues/171

sanderpick:
The idea here is to just ignore blocks when listing that have been flagged as "ignore".

fixes #172


wallet/thread/thread.go
```diff
@@ -98,7 +98,14 @@ func (t *Thread) Close() {
 // Blocks paginates blocks from the datastore
 func (t *Thread) Blocks(offsetId string, limit int, bType repo.BlockType) []repo.Block {
 	query := fmt.Sprintf("threadId='%s' and type=%d", t.Id, bType)
-	return t.blocks().List(offsetId, limit, query)
+	var filtered []repo.Block
+	for _, block := range t.blocks().List(offsetId, limit, query) {
+		ignored := t.blocks().GetByDataId(fmt.Sprintf("ignore-%s", block.Id))
```
sanderpick: here's the filtering...

andrewxhill: sweet

andrewxhill:
i've updated my mobile PR here https://github.com/textileio/textile-mobile/pull/270 with this as the fmwk

PR #220, opened 2018-08-05
feat(cmd): cmd helper, id from pk

sanderpick:


sanderpick:
Somewhat of a mishmash here... goes w/ https://github.com/textileio/textile-mobile/pull/310

mobile/mobile.go
```diff
@@ -541,8 +541,8 @@ func (m *Mobile) GetPhotoMetadata(id string) (string, error) {
 	}
 	meta, err := thrd.GetPhotoMetaData(id, block)
 	if err != nil {
-		log.Errorf("get photo meta data failed %s: %s", id, err)
-		return "", err
+		log.Warningf("get photo meta data failed %s: %s", id, err)
+		meta = &model.PhotoMetadata{}
```
sanderpick: here's where we just give back empty meta if it hasn't been found instead of an error - allowing the rest of a thread to be loaded

mobile/mobile.go
```diff
@@ -574,8 +574,8 @@ func (m *Mobile) getImageData(id string, path string, isThumb bool) (string, err
 	// get meta data for url type
 	meta, err := thrd.GetPhotoMetaData(id, block)
 	if err != nil {
-		log.Errorf("get photo meta data failed %s: %s", id, err)
-		return "", err
+		log.Warningf("get photo meta data failed %s: %s", id, err)
+		meta = &model.PhotoMetadata{}
```
sanderpick: same deal here, definitely want to just guess the image type of we can't find meta... have a feeling this is causing some `TextileImage` related crashes

wallet/thread/photos.go
```diff
@@ -124,6 +126,17 @@ func (t *Thread) HandleDataBlock(message *pb.Envelope, signed *pb.SignedThreadBl
 		}
 	}
 
+	// pin data
```
sanderpick: Pin the "off-chain" data... when we get to comments, will need a data type param on the message so we can determine what the dir structure should be, i.e., are we pinning a photo, comment, something else, etc.

wallet/wallet.go
```diff
@@ -465,6 +470,12 @@ func (w *Wallet) createIPFS(online bool) error {
 		return err
 	}
 
+	// determine routing
+	routing := core.DHTOption
+	if w.isMobile {
+		routing = core.DHTClientOption
```
sanderpick: Back to client DHT mode on mobile.

Issue #219, opened 2018-08-02
trouble building desktop client


whyrusleeping:
After a few dependency issues (the makefile doesnt fetch dependencies, needed a `go get` and `gx install`) I ran into this:
```
FATA[0082] bundling failed: bundling for environment darwin/amd64 failed: building failed: # github.com/textileio/textile-go/desktop
./bootstrap.go:24:11: unknown field 'Homepage' in struct literal of type bootstrap.Options
./bootstrap.go:25:17: unknown field 'MessageHandler' in struct literal of type bootstrap.Options
./bootstrap.go:26:9: cannot use start (type func(*astilectron.Astilectron, *astilectron.Window, *astilectron.Menu, *astilectron.Tray, *astilectron.Menu) error) as type bootstrap.OnWait in field value
./bootstrap.go:28:16: unknown field 'WindowOptions' in struct literal of type bootstrap.Options
: exit status 2  app_name=
make[1]: *** [build] Error 1
make: *** [build_desktop] Error 2
```

sanderpick:
Ohhh, yep, that's confusing... I recently reorganized the readme and moved the setup stuff to the bottom, below the desktop instructions :) I _think_ that looks like a version mismatch. We're using a combo of `dep` and `gx`, so I bundled the install into a single `npm run setup`. Three dependencies managers! So, if you do that (or just run `dep ensure` since you've got the gx deps) you should get the correct go-astilectron version, as specified in the `Gopkg.lock`.

Thanks for all the alpha testing :)

whyrusleeping:
Sweet, progress! Now it builds, but when running it i get this:
<img width="424" alt="screen shot 2018-08-02 at 9 41 15 am" src="https://user-images.githubusercontent.com/1243164/43597891-3bc17c6e-9638-11e8-98e9-c39a7ed5cd12.png">

carsonfarmer:
Hmm, we've had this issue before. I think it relates to this problem:
https://github.com/asticode/go-astilectron/issues/107 which required manually editing the astilectron version. Here's a similar issue: https://github.com/asticode/go-astilectron/issues/51 The fix there is actually better if it works.

carsonfarmer:
@whyrusleeping if you run the desktop app in dev mode with verbose outputs, that might confirm my suspicion above:
```
cd desktop/
go run *.go -v -d
```

carsonfarmer:
[...]

PR #224, opened 2018-08-09
Welcome messages, de-duplicated merges

sanderpick:
- Refactors how invites and joins work
  - move in-app and external invites off chain (reduce num. merges)
  - remove the single-echo from the thread originator when a new peer joins, replace with welcome messages from old peers, and a re-broadcasting on joins from the new peer after they've walked the chain and found more new peers
  - merge blocks with identical parents now hash identically as well (eek)
- Fixes up the desktop build process (#219) to use a vendored bundler, updates readme


wallet/thread/merges.go
```diff
@@ -1,15 +1,19 @@
...
+	"gx/ipfs/QmZoWKhxUmZ2seW4BzX6fJkNR8hh9PsGModr7q171yq2SS/go-libp2p-peer"
 	mh "gx/ipfs/QmZyZDi491cCNTLfAhwcaDii2Kg4pwKRkhqQzURGDvY6ua/go-multihash"
+	"sort"
 	"time"
 )
 
-// Merge adds an outgoing merge block, will only post to network if post is true
-func (t *Thread) Merge(head string, post bool) (mh.Multihash, error) {
+// Merge adds a merge block, which are kept local until subsequent updates, avoiding possibly endless echoes
+func (t *Thread) Merge(head string) (mh.Multihash, error) {
```
andrewxhill: do we hit similar problems as the repeater if we just have a crowded and busy room?

sanderpick: I think we might actually be able to post these now that they are hashed identical on each peer, but wanted to see how the current setup works... it seems to be sufficient and is def less noisy

PR #225, opened 2018-08-09
feat(profile): ipns avatar -> redirect to the linked url at /avatar_id

sanderpick:


andrewxhill:
sweet

sanderpick:
will work on any textile node gateway... desktop can use it's own gateway, but mobile should use the cafe

PR #227, opened 2018-08-09
Add more photo sizes, some GIF fixes

### LINKED ISSUE #170: New photo resizing
I'm not sure what the specific sizes are yet, but opening the ticket with some todos here,

- [ ] Thumbnail: big enough to represent a thumbnail in any UI (desktop, ipad, etc)
- [ ] Large image: big enough to be the full screen, full res image in any UI
- [ ] Original: unchanged

We can then optimize it over the network, grab thumbs always, grab Large at view time and sub-sample any other intermediate sizes we might want. Then just grab the original size in cases by request (e.g. backup sync on desktop)

sanderpick:
fixes #170, also properly resizes GIFs, and ensures that enlarged images arn't created during resizing

### DIFF
--- a/cafe/cafe.go
+++ b/cafe/cafe.go
@@ -5,7 +5,6 @@ import (
 	"github.com/gin-gonic/gin"
 	"github.com/op/go-logging"
 	cdao "github.com/textileio/textile-go/cafe/dao"
-	"github.com/textileio/textile-go/cafe/middleware"
 	"gx/ipfs/Qmb8jW1F6ZVyYPW1epc2GFRipmd3S8tJ48pZKBVPzVqj9T/go-ipfs/core"
 	"net/http"
 )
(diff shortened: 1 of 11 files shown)


sanderpick:
@asutula @andrewxhill, some small mobile API changes:
- `GetPhotoData` and `GetThumbData` is replaced with a single `GetPhotoData(id string, path string) (string, error)`, since you no longer need to tell the API if you want a thumbnail or a raw image (it will always return an encoded version). `path` is one of `thumb`, `small`, `medium`, `photo` (largest) `ImageData` no longer has `metadata` (AFAIK this object is only ever used by `TextileImage`, so should be ok)
- `PhotoMetadata` no longer has `username` (get it from the profile instead, where changes are reflected)
- `thumb` is used for avatars, might be ok enough for the grid
- `small` is about the size of the old `thumb`
- `medium` should be sufficient on most phones for a "full" size
- `photo` will return our current largest size and may be preferred on retina / hd devices / desktop
- `photo` will return our current large size

PR #229, opened 2018-08-10
Migration fix for fresh installs

sanderpick:
Also,
- won't auto-refresh messages on mobile
- breaks out mobile api into multiple files, was getting big

### DIFF
--- a/wallet/stats.go
+++ b/wallet/stats.go
@@ -1,0 +1,13 @@
+package wallet
+
+type Stats struct {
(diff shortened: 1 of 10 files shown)


wallet/stats.go
```diff
@@ -0,0 +1,13 @@
+package wallet
+
+type Stats struct {
```
sanderpick: was starting this... will finish in new PR

PR #234, opened 2018-08-14
feat(photos): index photo metadata

### LINKED ISSUE #233: Add photo aspect ratio to data blocks

sanderpick:
Did this one way (by adding the aspect ratio directly to the pb messages), then realized a much cleaner method: Index _all_ the metadata as a blob _on the way in_ by cating `/meta` when a photo is added to a thread and when it's received from another peer. This way, we can keep the same structure but only ever cat the data once per node.

fixes #233

### DIFF
--- a/cmd/photos.go
+++ b/cmd/photos.go
@@ -180,7 +180,7 @@ func GetPhoto(c *ishell.Context) {
 	c.Println(blue("saved to " + path))
 }
 
-func CatPhotoMetadata(c *ishell.Context) {
+func GetPhotoMetadata(c *ishell.Context) {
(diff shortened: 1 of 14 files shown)


sanderpick:
So, `metadata` is now fully available on the mobile `Photo` objects. So you'd get aspect ratio in the same way as you're doing now, w/h. If it's an old block, the metadata won't be available - you'll get back an empty metadata object, which will have w=0, h=0, so just check h != 0... maybe assume square in that case?

sanderpick:
Ended up not being too big of a change... took forever! G2G.

PR #236, opened 2018-08-14
getImageDataForMinWidth API

### LINKED ISSUE #232: GetImageDataForSize mobile API

sanderpick:
Getting image data was also improved by the switch to index meta data on the block... no longer need to decode the image data to determine the correct the data url prefix - can use the indexed encoding format.

fixes #232

### DIFF
--- a/mobile/photos.go
+++ b/mobile/photos.go
@@ -164,12 +164,19 @@ func (m *Mobile) GetPhotoData(id string, path string) (string, error) {
 		log.Errorf("get block data failed %s: %s", id, err)
 		return "", err
 	}
-	_, formatStr, err := image.DecodeConfig(bytes.NewReader(data))
+	var format string
+	meta, err := thrd.GetPhotoMetaData(id, block)
 	if err != nil {
-		log.Errorf("could not determine image format: %s", err)
-		return "", err
+		log.Warningf("get indexed photo meta data failed, decoding...")
+		_, format, err = image.DecodeConfig(bytes.NewReader(data))
+		if err != nil {
+			log.Errorf("could not determine image format: %s", err)
+			return "", err
+		}
+	} else {
+		format = meta.EncodingFormat
 	}
-	prefix := getImageDataURLPrefix(util.Format(formatStr))
+	prefix := getImageDataURLPrefix(util.Format(format))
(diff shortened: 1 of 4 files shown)


mobile/photos.go
```diff
@@ -164,17 +164,30 @@ func (m *Mobile) GetPhotoData(id string, path string) (string, error) {
...
+		_, format, err = image.DecodeConfig(bytes.NewReader(data))
+		if err != nil {
+			log.Errorf("could not determine image format: %s", err)
+			return "", err
+		}
+	} else {
+		format = meta.EncodingFormat
 	}
-	prefix := getImageDataURLPrefix(util.Format(formatStr))
+	prefix := getImageDataURLPrefix(util.Format(format))
```
asutula: So this prefix is like the `data:image/jpeg;` part?

sanderpick: yep

mobile/photos.go
```diff
@@ -164,17 +164,30 @@ func (m *Mobile) GetPhotoData(id string, path string) (string, error) {
...
 	encoded := libp2pc.ConfigEncodeKey(data)
 	img := &ImageData{Url: prefix + encoded}
 	return toJSON(img)
 }
 
+// GetPhotoDataForSize returns a data url of an image at or above requested size, or the next best option
+func (m *Mobile) GetPhotoDataForMinWidth(id string, minWidth int) (string, error) {
```
asutula: Awesome. Thinking through the consequences of only passing width instead of width and height... Close enough with just width?

sanderpick: That's what I was thinking... but maybe useful? Whatever you guys think. The other option for restricting via height I guess would be css + auto margin for centering

asutula:
This looks good. I feel good about the change, but thinking about it, I'm not 100% sure about how exactly we're going to implement the image data API, passing in the target width. Might take a little experimentation. For that reason, think maybe we should leave that API change in a branch while we experiment with integrating it into the app, just to be sure it's something we want to commit to 100%.

asutula:
Actually, I guess you're leaving the old API in place? If that's the case, then merge away.

sanderpick:
Yep, left it there. Cool, thanks!

PR #237, opened 2018-08-14
Unpin on ignore, expose to Mobile

### LINKED ISSUE #171: Unpin and ignore a photo in a thread

sanderpick:
When a block is ignored by adding an "ignore" block, the data the ignored block points to (if present) will be unpinned from the local node.

Remote unpinning is a little trickier... will kick that can down the road.

Mobile changes:
- `IgnorePhoto(blockId string) (string, error)` method, note that it needs `blockId`, not the photo's id because a photo may be in multiple threads
- `Photo` now has a `block_id` field, which would be needed if user wanted to ignore that photo

fixes #171

### DIFF
--- a/cmd/photos.go
+++ b/cmd/photos.go
@@ -238,9 +238,14 @@ func IgnorePhoto(c *ishell.Context) {
 	}
 	id := c.Args[0]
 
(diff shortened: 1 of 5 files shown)


sanderpick:
Done with this. Always more complicated than at first glance.

PR #258, opened 2018-08-21
Merge blocks can have dates too

sanderpick:
I had originally removed the date on the merge blocks because their hashes need to be deterministic from any peer that merges the same two parents. However, this kinda sucked cause it makes visualizing the update chain hard since the merge blocks were getting "dummy" dates (date at index time). Will also help when we revisit message ordering.


wallet/thread/merges.go
```diff
@@ -27,7 +29,33 @@ func (t *Thread) Merge(head string) (mh.Multihash, error) {
...
+	p1b := t.blocks().Get(header.Parents[0])
+	if p1b == nil {
+		return nil, errors.New("first merge parent not found")
+	}
+	p2b := t.blocks().Get(header.Parents[1])
+	if p2b == nil {
+		return nil, errors.New("second merge parent not found")
+	}
+	var date time.Time
+	if p1b.Date.Before(p2b.Date) {
```
sanderpick: Just pick the newest parent to use for date

wallet/thread/merges.go
```diff
@@ -27,7 +29,33 @@ func (t *Thread) Merge(head string) (mh.Multihash, error) {
...
+		date = p2b.Date
+	} else {
+		date = p1b.Date
+	}
+
+	// add a small amount to date to keep it ahead of both parents
+	date = date.Add(time.Millisecond)
```
sanderpick: and then we add a small amount to keep it ahead of both parents

PR #265, opened 2018-08-23
Delete notifications with related resources

### LINKED ISSUE #253: Cleanup notifications from deleted resources
- on thread removed, delete all notifications from that thread
- on peer left, delete their notifications
- on photo ignored, delete related notifications
- on device removed, delete related notifications

### LINKED ISSUE #260: Add a topic field
screen shot 2018-08-21 at 7 47 44 pm

type=2 notifications and other photo based ones, it would be ideal to include the photo id in the notification as well, making it so i can embed previews here and in local notifications (maybe)

sanderpick:
fixes #253 
fixes #260 

New `Notification` object looks like:
```
type Notification struct {
	Id            string           `json:"id"`
	Date          time.Time        `json:"date"`
	ActorId       string           `json:"actor_id"`                 // peer id
	ActorUsername string           `json:"actor_username,omitempty"` // peer username
	Subject       string           `json:"subject"`                  // thread name | device name
	SubjectId     string           `json:"subject_id"`               // thread id | device id
	BlockId       string           `json:"block_id,omitempty"`       // block id
	DataId        string           `json:"data_id,omitempty"`        // photo id, etc.
	Type          NotificationType `json:"type"`
	Body          string           `json:"body"`
	Read          bool             `json:"read"`
}
```

Refer to comments in ^ for client-side usage. Key points / changes:
- if you want a thread id, it will always be under `subject_id`
- if the notification pertains to a particular photo, it's id will be under `data_id`, which can be used to render a thumbnail
- `block_id` is used internally for deletes in some cases and for in-network invites (the invite id is the block id)
- the result of `AcceptThreadInviteViaNotification` is _no longer the thread id_ because it can be found on the actual notification under `subject_id`... the result is instead the join block's id, to be consistent w/ the rest of the api. 

Also, this fixes an issue with peer discovery that I ran into: If a peer ever needed to recover the thread, they'd have to find the original invite in order to discover the thread creator. So, this we just be sure to join our threads we create as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Until [Textile Photos](https://www.textile.photos) is ready for public release, 
 ## Install
 
 ```
-go get github.com/textileio/textile-go
+$ go get github.com/textileio/textile-go
 ```
 
 You'll need a few different tools here to get setup...
@@ -23,7 +23,7 @@ You'll need a few different tools here to get setup...
 Golang package manager:
 
 ```
-brew install dep
+$ brew install dep
 ```
 
 #### Install `gx`
@@ -31,8 +31,8 @@ brew install dep
 IPFS package manager:
 
 ```
-go get -u github.com/whyrusleeping/gx
-go get -u github.com/whyrusleeping/gx-go
+$ go get -u github.com/whyrusleeping/gx
+$ go get -u github.com/whyrusleeping/gx-go
 ```
 
 #### Install `node`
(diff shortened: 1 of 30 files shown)


andrewxhill:
might have a few Qs here. ActorId will always === a contact's id, yeah?

sanderpick:
Yep, `actorId` is always a peer id

PR #266, opened 2018-08-24
Block annotations (comments / likes)

### LINKED ISSUE #115: Comment on photos
This would be a dag structure like the photo directories, something like:

- a file called `parent` that contains the root cid of the target photo
- a file called `comment` that contains the actual comment text (much like `caption` file in the photo dag)
- a file called `meta` which is much like the photo `meta` file, containing, `created`, `username`, and `peer_id` fields.
- a file called `last` which contains the cid of the last comment OR photo update this peer made _in that room_, this way we still only need to republish HEAD update for each room

That covers the storage mechanism. We can do an indexing setup just like photos. So, a table in the SQLite db that has columns for the fields we want to index: `parent`, `created`, `username`, `peerId`.

### LINKED ISSUE #267: Ensure `refreshMessages` can only run once at a time
Same goes for publishing pointers.

sanderpick:
fixes #115
fixes #267 

Designed so that annotations are targeting blocks. So, for example, you could comment on / like somebody joining a thread: "jone joined" <- "mike and 3 others liked this".


sanderpick:
Ok, looking good here. Added mobile methods and types around:

- `func (m *Mobile) AddPhotoComment(blockId string, body string) (string, error)`
- `func (m *Mobile) AddPhotoLike(blockId string) (string, error)`

You'll have the `block_id` as it comes back with `GetPhotos`. 

- `func (m *Mobile) GetPhotos(offsetId string, limit int, threadId string) (string, error)` returns `Photo` types with added fields: `Comments` and `Likes`, which are arrays:

```
// Annotation represents common annotation fields
type Annotation struct {
	Id       string    `json:"id"`
	Date     time.Time `json:"date"`
	AuthorId string    `json:"author_id"`
	Username string    `json:"username,omitempty"`
}

// Comment is a simple wrapper around a comment block
type Comment struct {
	Annotation
	Body string `json:"body"`
}

// Like is a simple wrapper around a like block
type Like struct {
	Annotation
}
```

Starting to think more and more than we should remove the concept of return a list of "photos" in the mobile API, and replace it with just a list of the blocks, with the photo specific data joined into the block data... what we're doing now feels increasingly limiting. For example, we could have a generic `AddComment` to block method, instead of this photo specific one. I think it would be confusing to do that now. Anyway, good nuf for now.

Let's give it a whirl.

sanderpick:
Added mobile methods for ignoring comments and likes:

- `func (m *Mobile) IgnorePhotoComment(blockId string) (string, error)`
- `func (m *Mobile) IgnorePhotoLike(blockId string) (string, error)`

These just call a generic `ignoreBlock` method, but until we switch to a block-based mobile api, seems semantically more correct to have these treat the photo / comment / like as first class citizens.

andrewxhill:
okay, looks good. going to add to my PR and update the bridges, but not do any like/comment ui yet

PR #273, opened 2018-08-28
A couple fixes

### LINKED ISSUE #269: getTokens optional return value
Just an idea, let's discuss it... Right now `getTokens` throws an sql error if it's called when there are no tokens to get. Should it just return a null value instead or a more meaningful error?

### LINKED ISSUE #270: Mobile: Photo comment array coming back null
Looks like it's null when empty even though the API says it can't be null. Should be an empty array. Might as well check Likes while we're at it.

### LINKED ISSUE #271: Add caching header to cafe

sanderpick:
- Fixes #269: Refactored datastore profile methods to not throw sql "no row" errors. This means that on mobile, `GetTokens`, `GetUsername`, and `GetAvatarId` will return empty strings if nothing is found and NO error. (we can't return `nil` because `gobind` doesn't support string pointers.
- Fixes #271 Adds caching to avatars. The mechanism no longer uses a redirect, instead loads content behind the `avatar_id` directly. Expiration set at 2 days.
- Fixes #270 `Comments` and `Likes` will now be empty arrays (not null)
- Removes `AddThreadWithMnemonic`, it's not useful anymore and likely making creating threads slower from mobile, which was wrapping that method. Mobile now uses the normal create thread method, meaning it's signature doesn't have the string param `mnemonic` anymore
- Lastly, adds the actual comment to comment notifications

### DIFF
--- a/net/service/handlers.go
+++ b/net/service/handlers.go
@@ -3,6 +3,7 @@ package service
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes"
 	"github.com/golang/protobuf/ptypes/any"
@@ -330,7 +331,11 @@ func (s *TextileService) handleThreadAnnotation(pid peer.ID, pmes *pb.Envelope, 
 		if err != nil {
 			return nil, err
 		}
-		notification.Body = "commented on " + target
+		body, err := thrd.Decrypt(annotation.CaptionCipher)
+		if err != nil {
+			return nil, err
+		}
+		notification.Body = fmt.Sprintf("commented on %s: \"%s\"", target, string(body))
 	case pb.ThreadAnnotation_LIKE:
 		notification, err = buildNotification(thrd.PrivKey, annotation.Header, repo.LikeAddedNotification)
 		if err != nil {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,21 +85,54 @@ jobs:
           CAFE_REFERRAL_KEY: woohoo!
 
   build-cli:
-    docker:
-    - image: textile/builder:1.10.3
-    working_directory: /go/src/github.com/textileio/textile-go
+    macos:
+      xcode: "9.4.0"
+    environment:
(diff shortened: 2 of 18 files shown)


asutula:
I can't decide if we should commit to returning empty strings for those methods that previously threw sql errors. Even though the errors don't feel ideal, I'm wondering if that is the more idiomatic solution, especially if we want to make the generated frameworks usable in Java/Swift/Obj C. Thoughts?

asutula:
Leaning towards just making the error more meaningful.

sanderpick:
I'd prefer to stick with the empty string for now. It's the best we can do for "doesn't exist", and not existing is not an error state here. Also in Go, empty strings are just the "0 value" of a string, since optionals aren't a thing. Another option is to make the bridges take the string and convert to whatever language specific null value it wants.

sanderpick:
We'd return a `nil` byte slice. A string here is more consistent with the current API.

asutula:
cool

sanderpick:
We could also return an empty object in JSON `"{}"`... maybe that's more consistent?

sanderpick:
(since the expected result is an object)

andrewxhill:
Did you list exhaustively the methods that now could return ''? Or should I be looking into getId, getPublicKey etc?

sanderpick:
I think `GetUsername` is the only other method that could return `""` as a non-error state, as in, while at the same time as `err == nil`

net/service/handlers.go
```diff
@@ -330,7 +331,11 @@ func (s *TextileService) handleThreadAnnotation(pid peer.ID, pmes *pb.Envelope,
 		if err != nil {
 			return nil, err
 		}
-		notification.Body = "commented on " + target
+		body, err := thrd.Decrypt(annotation.CaptionCipher)
+		if err != nil {
+			return nil, err
+		}
+		notification.Body = fmt.Sprintf("commented on %s: \"%s\"", target, string(body))
```
andrewxhill: feels a bit like it would be nicer to have these split out so i can easily rearrange the text based on the type of display (app notification or feed row). fine for now

PR #280, opened 2018-09-07
feat(go-mobile): add mobile npm package

sanderpick:
Adds `@textile/go-mobile` npm package to the `mobile` directory.


mobile/scripts/fetch_release.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+DIR="$(dirname "$(stat -f "$0")")"
+IOS_DIR="${DIR}/../ios"
+ANDROID_DIR="${DIR}/../android"
+
+# fetch iOS framework
+mkdir -p ${IOS_DIR}
+wget https://github.com/textileio/textile-go/releases/download/v0.1.4/textile-go_v0.1.4_ios-framework.tar.gz
```
asutula: It's fine all these version numbers are hard coded for now?

asutula: Seems like the version string could be passed as an arg from package.json to this script

sanderpick: it's in a half-baked state now where this package version doesn't match the go release version (cause i had already published 0.1.4)... so, next release can clean all that up... ideally, in CI we'd bump both versions together, then auto-publish this... something like pass version to script, like you suggest. fine with it like this until next time around

sanderpick:
Slack down... but good 2 go here.

sanderpick:
`@textile/go-mobile@0.1.5-rc.7` is latest w/ curl

mobile/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "keywords": [
+    "textile",
+    "decentralized",
+    "mobile"
+  ],
+  "files": [
+    "scripts"
+  ],
+  "scripts": {
+    "postinstall": "./scripts/fetch_release.sh 0.1.5"
```
asutula: We should just use `$npm_package_version` here so you don't ever have to update that

sanderpick: oh nice, didnt know about that

asutula: Think it works, def test it haha

sanderpick: yep, works

PR #281, opened 2018-09-07
cant use `stat` on linux like on macos

sanderpick:


mobile/package.json
```diff
@@ -18,6 +18,6 @@
     "scripts"
   ],
   "scripts": {
-    "postinstall": "./scripts/fetch_release.sh $npm_package_version"
+    "postinstall": "./scripts/fetch_release.sh 0.1.5"
```
asutula: Intentional 0.1.5?

sanderpick: Yeah - I don't wanna have to release an entire new go lib for every little fix here. feels like releases should be more "major". Think we'll have to put some thought into versioning (see slack chatter)

Issue #285, opened 2018-09-11
Profiles/identity for resilient backup & recovery


sanderpick:
The current user "profile" is half-baked and split between the cafe and IPNS. It looks like...

- **Username + Password** for cafe registration (stored by a cafe, pass salted+hashed)

The _only_ reason to require a username + pass combo is for handling cafe pin requests. In practice, this means storing a valid JWT as the user's "session".

- **Email Address** (stored by a cafe)

Used for beta feedback. Would be nice to remove.

- **Avatar** (stored via IPNS)

The main problem is that since the IPNS `id` behind a profile is the actual peer `id` of the local IPFS node, "login" means something like logging into that specific IPFS peer, which won't ever make sense. Peers must not be duplicated on the network and trying to manage peer existence across devices is not practical / impossible.

There are other issues w/ the current setup, but I think this will be more clear if we start with use cases:

- Portability: real login / logout from a "device"
- Backup: in a way, same as above, but with an emphasis on the ability to do a full data recovery from the network alone, i.e., without any existing devices
- Discoverability: search and be found on your terms
- Multiple profile support: as "profile" context is part and parcel with its contents, we want the spec to include the ability to recover multiple profiles from the same key / mnemonic

The proposal is a DAG with a mnemonic phrase at the root, profiles at the first level, and IPFS peers at the third level:

```
MnemonicPhrase
├── PrivateKey
└── PublicKey
    ├── ProfileA
    │   ├── Peers
    │   │   ├── A
    │   │   ├── B
    │   │   └── C
    │   ├── PrivateKey
    │   └── PublicKey
    │       ├── AvatarUri_plaintext
    │       ├── EmailAddress_ciphertext
    │       ├── PhoneNumber_ciphertext
    │       ├── Threads
    │       │   ├── A
    │       │   │   ├── HEAD_ciphertext
    │       │   │   └── PrivateKey_ciphertext
    │       │   ├── B
    │       │   │   ├── HEAD_ciphertext
    │       │   │   └── PrivateKey_ciphertext
    │       │   └── C
    │       │       ├── HEAD_ciphertext
    │       │       └── PrivateKey_ciphertext
    │       └── Username_plaintext
    └── ProfileB
        ├── Peers
        │   ├── A
        │   ├── B
        │   └── C
        ├── PrivateKey
        └── PublicKey
            ├── AvatarUri_plaintext
            ├── EmailAddress_ciphertext
            ├── PhoneNumber_ciphertext
            ├── Threads
[...]

PR #288, opened 2018-09-14
feat(cafe): adds api v1 w/ challenge based token issueing

sanderpick:
Some of the cafe work for #285, removing the need for username+passwords. Authentication protocol looks like:

345px-nonce-cnonce-uml svg

Except that "username" in the diagram is the profile's `pk`, and the hash function is replaced by an actual signature: `sig(snonce + cnonce)`. We still return a HS256 (symmetric) JWT as the "session".


sanderpick:
Fixed up tests, have a look-see.

PR #298, opened 2018-09-18
fix(cafe): limit password to 24 chars

sanderpick:
long passwords take forever to process


carsonfarmer:
super minor change, once tests pass, should be gtg :) thanks for the find and fix @sanderpick!

PR #299, opened 2018-09-22
BIP32-based wallet refactor

### LINKED ISSUE #292: Re-investigate HD Wallet for deriving profile keys
(See project description for visual)

Re: the top level keypair, or "root" key pair, the intended use case is to allow a user to recover (or just view for login) all their profiles from a single mnemonic phrase. One key feature here is that a Textile app _need not handle the mnemonic_, but can just take a profile key. So, if we're simply using and encrypted blob of profile keys under the root key's IPNS address, we run into a fatal flaw... **there won't be a peer responsible for IPNS re-publishing**.

If we instead use an HD Wallet, the keys will always be available because they're actually contained in the mnemonic, so to speak... no need to store elsewhere on IPNS, a cafe, etc. This would also lead us nicely down a path where we could actually see an integration with hard wallets like Ledger.

To investigate:
- Explore the correct key derivation path
- Can we use Ed25519 keys at the leaves? (maybe already the case, need to play with https://github.com/tyler-smith/go-bip32)

We'd be setup for a standalone app, _Textile Wallet_, which can generate a new mnemonic, view all existing profiles (walk the derivation path, discovering content from IPNS profiles), and let you create new profiles. So then to login into a Textile app, you just copy a profile key to clipboard in Textile Wallet, open the target app, paste the key... you're in and the app doesn't need to know your root key.

### LINKED ISSUE #295: Migrate local login / session to challenge based flow

### LINKED ISSUE #296: Refactor node creation to begin with a profile key

sanderpick:
fixes #292
fixes #295 
fixes #296


sanderpick:
Refactored the internal life cycle to handle the new "account" concept. As a result, the command line needed some big changes. Now it's sub command based:

```
Usage:
  textile [OPTIONS] <command>

Application Options:
  -r, --repo-dir=     specify a custom repository path

Log Options:
  -l, --log-level=    set the logging level [debug, info, notice, warning, error, critical] (default: debug)
  -n, --no-log-files  do not save logs on disk

Help Options:
  -h, --help          Show this help message

Available commands:
  daemon   Start a node daemon
  init     Init the node repo
  shell    Start a node shell
  version  Print version and exit
```

Adding `--help` to sub commands prints detailed options.

To get started, you'd init a repo (as w/ `ipfs`) like so:
```
textile init -s STqdutoFiPV9Nzh96L7bvSZNzrHkQiB2phDW1ySG9Na1kkGE
```

Then run the daemon:
```
textile daemon
```
Or the interactive shell:
```
textile shell
```

One big benefit here is that you are no longer able to overwrite your repo 💯

sanderpick:
One more example... `textile shell --help`:

```
Usage:
  textile [OPTIONS] shell [shell-OPTIONS]

Start an interactive node shell session.

Application Options:
  -r, --repo-dir=              specify a custom repository path

Log Options:
  -l, --log-level=             set the logging level [debug, info, notice, warning, error, critical] (default: debug)
  -n, --no-log-files           do not save logs on disk

Help Options:
  -h, --help                   Show this help message

[shell command options]

    Gateway Options:
      -g, --gateway-bind-addr= set the gateway address (default: 127.0.0.1:random)

    Cafe Options:
      -c, --cafe=              cafe host address
          --cafe-bind-addr=    set the cafe address
          --cafe-db-hosts=     set the cafe mongo db hosts uri
          --cafe-db-name=      set the cafe mongo db name
          --cafe-db-user=      set the cafe mongo db user
          --cafe-db-password=  set the cafe mongo db user password
          --cafe-db-tls        use TLS for the cafe mongo db connection
          --cafe-token-secret= set the cafe token secret
          --cafe-referral-key= set the cafe referral key
```

sanderpick:
Further cleanup for the `wallet` command:

```
$ textile --help
Usage:
  textile [OPTIONS] <command>

Help Options:
  -h, --help  Show this help message

Available commands:
  daemon   Start a node daemon
  init     Init the node repo and exit
  shell    Start a node shell
  version  Print version and exit
  wallet   Manage a wallet of accounts
```

```
$ textile wallet --help
Usage:
  textile [OPTIONS] wallet <accounts | init>

Initialize a new wallet, or view accounts from an existing wallet.

Help Options:
  -h, --help      Show this help message

Available commands:
  accounts
  init
```

sanderpick:
Tests passing on this, finally. Plenty more to do, but this PR is cooked. Can now move to smaller changes. Can I haz review approval, anyone?

core/account.go
```diff
@@ -0,0 +1,54 @@
+package core
```
carsonfarmer: Ok, this module is quite nice!

PR #300, opened 2018-10-02
libp2p CafeService for thread backup and offline inboxing

sanderpick:
- Updates `go-ipfs` to 0.4.17
- Improved modular libp2p service layer as `Service`
- Refactored thread handlers as `ThreadService`, riding on `Service`
- Adds `CafeService`, riding on `Service`, with thread backup methods + handlers.
- Fully encrypt thread blocks (major change to thread structure for better privacy)


sanderpick:
Tricky publishing under a different name (new account key)... finally realized you have to add the libp2p crypto style priv key to the ipfs repo's keystore. 

Realizing that it's gonna be pretty tough to keep this thread list in sync between peers. Imagine, who gets to update? Which write wins? Does each peer need to first get the latest, check timestamp of what it knows to be the latest update in a thread, and only update if it's newer? Sounds like a nightmare. I think we need to namespace the thread list under peer id, instead of having an additional list of peer ids at the same level (as seen in the project description). This way, the nastiness is pushed to the less frequently run recovery mechanism... i.e., use the most up-to-date (recent) thread state from all the peers.

sanderpick:
Good to go for a look + merge. Couple outstanding issues before a pre-release (see project), but they are isolated.

sanderpick:
doin it

PR #331, opened 2018-10-25
Slack integration

carsonfarmer:
This is a first 'phase' towards adding 'integrations' to Textile Photos. Here, I've added some optional cli arguments to `shell` and `daemon` modes that allows a Cafe to automatically post any photos added to any Thread the Cafe has joined to a specified webhook. Currently, we only support Slack incoming webhooks, but I could see us adding more generic support later. There is a longer roadmap for this which we will outline elsewhere.

Currently, there are some hardcoded parameters and urls that should eventually become configurable. But this should be more than sufficient for testing and initial trials.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,15 @@ See [go-astilectron-bundler](https://github.com/asticode/go-astilectron-bundler)
 
 Note: Because `cgo` is required, you'll need to setup a proper C toolchain for cross-OS-compilation.
 
+## Webhook Integration
+
+After building the CLI/daemon, run either `shell` or `daemon` with the optional `--slack-hook` (or `-s`) and/or `--welcome-image` flags. Adding a welcome image will automatically post a welcome message and image to all Threads that the textile daemon has joined. While this is setup to support Slack webhooks, it should work for any webhook that can mimic [Slack's API](https://api.slack.com/incoming-webhooks).
+
+```
+textile daemon -s https://hooks.slack.com/services/blah/blah --welcome-image ./welcome.png
+```
+
+
 ## Contributing
 
 The easiest way to write a valid commit message is to use the `npm` script:
--- a/textile.go
+++ b/textile.go
@@ -1,8 +1,20 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	icore "gx/ipfs/Qmb8jW1F6ZVyYPW1epc2GFRipmd3S8tJ48pZKBVPzVqj9T/go-ipfs/core"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+
 	"github.com/fatih/color"
 	"github.com/jessevdk/go-flags"
 	"github.com/mitchellh/go-homedir"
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -167,6 +167,15 @@ See [go-astilectron-bundler](https://github.com/asticode/go-astilectron-bundler)
 
 Note: Because `cgo` is required, you'll need to setup a proper C toolchain for cross-OS-compilation.
 
+## Webhook Integration
```
carsonfarmer: Minimal docs. Maybe we don't even want to document this just yet?

textile.go
```diff
@@ -55,6 +62,10 @@ type CafeOptions struct {
 	DBTLS       bool   `long:"cafe-db-tls" description:"use TLS for the cafe mongo db connection"`
 	TokenSecret string `long:"cafe-token-secret" description:"set the cafe token secret"`
 	ReferralKey string `long:"cafe-referral-key" description:"set the cafe referral key"`
+
+	// Slack hook url settings
```
carsonfarmer: Added these to CafeOptions because then it'll work nicely in both shell and daemon modes. _Could_ do this as its own top-level group, but seems to fit here conceptually.

textile.go
```diff
@@ -98,6 +109,28 @@ type ShellCommand struct {
 	Cafe     CafeOptions    `group:"Cafe Options"`
 }
 
+// Payload is a Slack message payload
+type Payload struct {
```
carsonfarmer: These are used when creating the urlencoded parameters for the wehbook. This is only a subset of the keys supported by Slack, but its all we're using, so might as well keep it simple. Possible improvement, because we use defaults, is to create a constructor with the default values in it.

textile.go
```diff
@@ -327,6 +362,103 @@ func getRepoPath(repoPath string) (string, error) {
 	return repoPath, nil
 }
 
+func threadIntegrationUpdate(update thread.Update, cafeURL string, hookURL string) error {
```
carsonfarmer: Setup integration, lots of hardcoded things

textile.go
```diff
@@ -327,6 +362,103 @@ func getRepoPath(repoPath string) (string, error) {
...
+		Text:        fmt.Sprintf("added a photo to *'%s*'", update.ThreadName),
+		Markdown:    true,
+		IconURL:     fmt.Sprintf("%s/ipns/%s/avatar", cafeURL, authorID.Pretty()),
+		Attachments: []*Attachment{attachment},
+	}
+
+	return postPayloadToHook(payload, hookURL)
+}
+
+func postPayloadToHook(payload *Payload, hookURL string) error {
```
carsonfarmer: Separated out in case we want to do this elsewhere

textile.go
```diff
@@ -422,6 +554,15 @@ func startNode(cafeOpts CafeOptions, gatewayOpts GatewayOptions) error {
 				if !ok {
 					return
 				}
+				if cafeOpts.SlackHook != "" && cafeOpts.Addr != "" {
```
carsonfarmer: Very easy to add here, limited to only photos, but we could defs supports comments, etc. if we wanted.

textile.go
```diff
@@ -502,6 +643,36 @@ func printSplashScreen(cafeOpts CafeOptions, daemon bool) {
 	}
 }
 
+func welcomeSlackMessage(cafeOpts CafeOptions, daemon bool) {
```
carsonfarmer: Very simple message to thread

carsonfarmer:
Closing as we'll keep this in a separate branch for now. Refer to this PR for comments etc.

PR #351, opened 2018-11-10
feat(desktop): adding tray util for wallet sync

andrewxhill:
moves desktop ui out and just adds basic sync tool


andrewxhill:
this is going to be so easy i think. biggest unknown for me is what to do with the selected target folder for photo syncing, but i'm sure you guys know a way there.

PR #352, opened 2018-11-14
Adds steaming/updates endpoint to cafe to support real-time updates on clients

carsonfarmer:
To support an arbitrary number of listeners to thread updates, also adds a broadcasting system for
'channels'. Currently, only threadUpdates uses this system, but easy to add others.


carsonfarmer:
Note, currently the stream returns the block information. This isn't all that useful from a client perspective. So once we have a nicer API for file objects, we should move to use that.

carsonfarmer:
No idea what happened with that failed test @sanderpick will try to take a look later...

core/api.go
```diff
@@ -64,6 +66,14 @@ func (a *api) Start() {
 		g.Writer.WriteHeader(http.StatusNoContent)
 	})
 
+	// Allows all origins
+	// TODO: Do not use this in production, needs to be configurable
+	config := cors.DefaultConfig()
```
carsonfarmer: This should probably really be part of a different PR... required to use any of our api commands in CORS context.

core/api.go
```diff
@@ -88,8 +98,9 @@ func (a *api) Start() {
 		threads.GET("/:id", a.getThreads)
 		threads.DELETE("/:id", a.rmThreads)
 		threads.POST("/:id/files", a.addThreadFiles)
+		threads.GET("/:id/updates", a.streamThreads)
```
carsonfarmer: does this seem like a reasonable endpoint?

core/api.go
```diff
@@ -88,8 +98,9 @@ func (a *api) Start() {
...
 
-		files := v0.Group("/files")
+    files := v0.Group("/files")
```
carsonfarmer: auto-formatting seems off here?

core/api.go
```diff
@@ -135,12 +146,13 @@ func (a *api) Start() {
 
 // Stop stops the http api
 func (a *api) Stop() error {
-	ctx, cancel := context.WithCancel(context.Background())
+	// Use timeout to force a deadline
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
```
carsonfarmer: useful for channels that take too long to close

core/api_threads.go
```diff
@@ -121,3 +122,86 @@ func (a *api) rmThreads(g *gin.Context) {
...
+
+func addBlockInfo(a *api, update ThreadUpdate) (ThreadUpdate, error) {
+	block := update.Block
+	username := a.node.ContactUsername(block.AuthorId)
+
+	var info interface{}
+	switch update.Block.Type {
+	case repo.FilesBlock:
+		info, _ = a.node.threadFile(&update.Block)
+	case repo.CommentBlock:
```
carsonfarmer: seems overly verbose? better way to do this?

core/broadcaster.go
```diff
@@ -0,0 +1,106 @@
+// Package core/broadcast implements multi-listener broadcast channels.
```
carsonfarmer: Left this in core. Should it be its own package?

core/main.go
```diff
@@ -265,7 +265,7 @@ func (t *Textile) Start() error {
 
 	t.online = make(chan struct{})
 	t.updates = make(chan Update, 10)
-	t.threadUpdates = make(chan ThreadUpdate, 10)
+	// t.threadUpdates = make(chan ThreadUpdate, 10)
```
carsonfarmer: can be removed, left for review... should be gone. broadcaster is lazily instantiated, so don't need this

core/main.go
```diff
@@ -448,8 +448,8 @@ func (t *Textile) UpdateCh() <-chan Update {
 }
 
 // ThreadUpdateCh returns the thread update channel
-func (t *Textile) ThreadUpdateCh() <-chan ThreadUpdate {
-	return t.threadUpdates
+func (t *Textile) ThreadUpdateCh() *Listener {
```
carsonfarmer: this now returns a listener. should the method be renamed to reflect this change?

core/thread.go
```diff
@@ -40,9 +40,10 @@ var ErrBlockNotFile = errors.New("block is not a file")
 
 // ThreadUpdate is used to notify listeners about updates in a thread
 type ThreadUpdate struct {
-	Block      repo.Block `json:"block"`
-	ThreadId   string     `json:"thread_id"`
-	ThreadName string     `json:"thread_name"`
+	Block      repo.Block  `json:"block"`
+	ThreadId   string      `json:"thread_id"`
+	ThreadName string      `json:"thread_name"`
+	Info       interface{} `json:"info,omitempty"`
```
carsonfarmer: info contains useful, human-readable, thread update information

core/threads_files.go
```diff
@@ -43,6 +43,20 @@ type ThreadLikeInfo struct {
 	Username string    `json:"username,omitempty"`
 }
 
+type ThreadJoinInfo struct {
```
carsonfarmer: copy-paste here... do we need these? reuse other interface perhaps?

mill/image_resize.go
```diff
@@ -133,8 +133,15 @@ func encodeImage(reader io.Reader, format Format, width int, quality int) (*byte
 		}
 
 		resized := imaging.Resize(img, width, 0, imaging.Lanczos)
-		if err = jpeg.Encode(buff, resized, &jpeg.Options{Quality: quality}); err != nil {
-			return nil, nil, err
+
+		if format == PNG {
```
carsonfarmer: quick fix to make sure we encode pngs correctly

mobile/mobile.go
```diff
@@ -177,10 +177,11 @@ func (m *Mobile) Start() error {
 		}()
 
 		// subscribe to thread updates
+		listener := m.node.ThreadUpdateCh()
```
carsonfarmer: new thread update api usage here as well

textile.go
```diff
@@ -394,24 +394,30 @@ func startNode() error {
 		}
 	}()
 
-	// subscribe to thread updates
+	// Subscribe to thread updates
+	listener := node.ThreadUpdateCh()
```
carsonfarmer: also here

carsonfarmer:
Yeh hmmm, re-ran them all locally, looks good here?

cmd/threads.go
```diff
@@ -300,6 +304,72 @@ func callRmThreads(args []string) error {
...
+
+func (x *streamThreadsCmd) Short() string {
+	return "Steam thread updates"
+}
+
+func (x *streamThreadsCmd) Long() string {
+	return `
+Streams all thread update types.
+Use the --events option to emit Server-Sent Events (SSEvent), otherwise, emit JSON responses. 
+SSEvents enable browsers/clients to consume the stream using EventSource.
```
sanderpick: hell yeah

core/api.go
sanderpick: cool, can you open an issue to make this part of the textile config file?

core/api.go
sanderpick: lgtm

core/api.go
sanderpick: perf

core/api.go
sanderpick: huh, yeah

core/contacts.go
```diff
@@ -75,7 +75,7 @@ func (t *Textile) ContactUsername(id string) string {
 	if id == "" {
 		return ""
 	}
-	username = id[len(id)-7:]
+	username = id[len(id)-8:]
```
sanderpick: it's 7 everywhere else... let's keep that

core/main.go
sanderpick: nice! kill it

core/main.go
sanderpick: sounds right... `GetThreadUpdateListener`?

core/broadcaster.go
sanderpick: cool stuff. yeah, seems like it could be it's own package

mill/image_resize.go
sanderpick: sweet

core/threads_files.go
sanderpick: yeah, hmmm. I see what you mean. I just added username to the BlockInfo object, which is available via Textile.BlockInfo(blockId)... how about just replaceing `Block` on ThreadUpdate w/ `BlockInfo`, and then only in the case where you want to inflate the files update info tack on the `Info` field?

core/api_threads.go
sanderpick: hhmmm, yeah. see comment below about `BlockInfo`

sanderpick:
failure is this annoying map concurrent read thing, happens once in awhile (only in tests w/ multiple nodes trying to access the ipfs go-logging module)

core/api.go
carsonfarmer: added

core/api.go
carsonfarmer: fixed in upcoming push

core/contacts.go
carsonfarmer: oh hah, just realized what i did there... nv, reverting 😬

carsonfarmer:
aside from that annoying test issue, this should be GTG now sander

core/thread.go
```diff
@@ -441,7 +442,18 @@ func (t *Thread) indexBlock(commit *commitResult, blockType repo.BlockType, targ
...
-	t.pushUpdate(*index)
+	info := &BlockInfo{
+		Id:       index.Id,
+		ThreadId: index.ThreadId,
+		AuthorId: index.AuthorId,
+		Type:     index.Type.Description(),
+		Date:     index.Date,
+		Parents:  index.Parents,
+		Target:   index.Target,
+		Body:     index.Body,
```
sanderpick: if you do `t.BlockInfo(index.Id)` you'll get the info w/ `Username` hydrated

core/api_threads.go
```diff
@@ -121,3 +122,65 @@ func (a *api) rmThreads(g *gin.Context) {
...
+func addBlockInfo(a *api, update ThreadUpdate) (ThreadUpdate, error) {
+	block := update.Block
+	username := a.node.ContactUsername(block.AuthorId)
+
+	var info interface{}
+	switch update.Block.Type {
+	case "FILES":
+		info, _ = a.node.ThreadFile(update.Block.Id)
+	case "COMMENT":
+		info = ThreadCommentInfo{
```
sanderpick: w/ username on block info, you could kill this too.. redundant

core/api_threads.go
```diff
@@ -121,3 +122,65 @@ func (a *api) rmThreads(g *gin.Context) {
...
+}
+
+func addBlockInfo(a *api, update ThreadUpdate) (ThreadUpdate, error) {
+	block := update.Block
+	username := a.node.ContactUsername(block.AuthorId)
+
+	var info interface{}
+	switch update.Block.Type {
+	case "FILES":
+		info, _ = a.node.ThreadFile(update.Block.Id)
```
sanderpick: you could `log.Errorf("error getting thread file: %s", err)` here

core/thread.go
carsonfarmer: didn't go this route, but as per our slack convo, merged master and added username to block info

carsonfarmer:
pushed up a sneaky fix, will merge when done checks

PR #356, opened 2018-11-21
Improved mobile add files flow

sanderpick:


sanderpick:
@asutula I realized the way has originally smashed in the "async + proto" test was gonna be pretty cumbersome to try out. This should be much nicer. You can see usage in `mobile_test.go`.

mobile/mobile_test.go
```diff
@@ -211,23 +208,26 @@ func TestMobile_PrepareFiles(t *testing.T) {
 		t.Errorf("prepare files failed: %s", err)
 		return
 	}
-	dir := core.Directory{}
-	if err := json.Unmarshal([]byte(res), &dir); err != nil {
+	pre := new(pb.MobilePreparedFiles)
```
sanderpick: this is the result of calling `PrepareFiles` (sync and async)

mobile/mobile_test.go
```diff
@@ -211,23 +208,26 @@ func TestMobile_PrepareFiles(t *testing.T) {
...
+	if err := proto.Unmarshal(res, pre); err != nil {
 		t.Error(err)
 		return
 	}
-	if len(dir) != 6 {
+	if len(pre.Dir.Files) != 6 {
 		t.Error("wrong number of files")
 	}
-	prepared = res
+	dir, err = proto.Marshal(pre.Dir)
```
sanderpick: `AddThreadFile` takes marshalled `Directory` bytes, which is one field from the prepared result.

schema/main.go
```diff
@@ -19,6 +19,9 @@ var ErrLinkOrderNotSolvable = errors.New("link order is not solvable")
 // FileTag indicates the link should "use" the input file as source
 const FileTag = ":file"
 
+// SingleFileTag is a magic key indicating that a directory is actually a single file
+const SingleFileTag = ":single"
```
sanderpick: @carsonfarmer the file vs. dir thing was getting pretty annoying. so, the api now _always_ expects a dir of files. If the schema demands a single file (no links), it will com in with a single link name `:single`, which is then parsed out.

PR #357, opened 2018-11-22
Update go-ipfs to 0.4.18

sanderpick:
Had to make a bunch of little tweaks... seems to be working though.


crypto/asymmetric.go
```diff
@@ -63,6 +66,21 @@ func encryptCurve25519(pubKey *libp2pc.Ed25519PublicKey, bytes []byte) ([]byte,
 	return ciphertext, nil
 }
 
+func publicToCurve25519(k *libp2pc.Ed25519PublicKey) (*[32]byte, error) {
```
sanderpick: libp2p-crypto remove this `ToCurve25519` method and it's private key variant... but mentioned in the commit msg that can easily be pulled back in. Doing that here.

crypto/asymmetric.go
```diff
@@ -96,7 +118,18 @@ func decryptCurve25519(privKey *libp2pc.Ed25519PrivateKey, ciphertext []byte) ([
 	return plaintext, nil
 }
 
-// verify verifies a signature
+func privateToCurve25519(k *libp2pc.Ed25519PrivateKey) (*[32]byte, error) {
```
sanderpick: and here

PR #361, opened 2018-11-24
File unpinning and de-indexing

sanderpick:
Painstakingly handles file unpins and de-indexes. The only way to do this was to add an extra field on `File` which is a list of the targets (DAG top level hashes) that the file belongs to. We unpin / de-index a file index and it's data (if pinned by schema), if 1) it's not part of another file structure, and 2) it's parent target is not part of another block.


GitCop:

 * Commit: 70d919d76a0e632b7f4e1da53fcb1316e7365e6e
  * Commits must be in the following format: **%{type}(%{scope}): %{description}**

Guidelines are available at https://github.com/textileio/textile-go

---

PR #365, opened 2018-11-25
files: adds directory handling to cli

sanderpick:
Pretty nice to ingest an entire directory... need to add batching though.


sanderpick:
added the batching

PR #367, opened 2018-11-25
fix infinite broadcast listener loops

sanderpick:
You might have noticed @carsonfarmer - the textile cmd currently consumes >100% CPU. I narrowed it down to the broadcast commits... finally noticed you have a `default` case in an infinite for loop that is spinning its wheels to no end.


GitCop:

 * Commit: cba332390123e38a57c9b4d3ad820d0952ac6e3b



---

broadcast/main.go
```diff
@@ -27,12 +33,12 @@ type Listener struct {
 }
 
 // Send broadcasts a message to the channel.
-// Sending on a closed channel causes a runtime panic.
 func (b *Broadcaster) Send(v interface{}) {
 	b.m.Lock()
 	defer b.m.Unlock()
 	if b.closed {
-		panic("broadcast: send after close")
+		log.Warning("send on closed channel")
```
sanderpick: may as well not panic here

core/main.go
```diff
@@ -197,7 +197,12 @@ func NewTextile(conf RunConfig) (*Textile, error) {
 	// force open the repo and datastore
 	removeLocks(conf.RepoPath)
 
-	node := &Textile{repoPath: conf.RepoPath}
+	node := &Textile{
+		repoPath:      conf.RepoPath,
+		updates:       make(chan Update, 10),
+		threadUpdates: broadcast.NewBroadcaster(10),
+		notifications: make(chan repo.Notification, 10),
```
sanderpick: initing all these channels on node create instead of start

core/main.go
```diff
@@ -385,15 +388,18 @@ func (t *Textile) Stop() error {
...
-	t.threadUpdates.Close()
-	close(t.notifications)
-
 	log.Info("node is stopped")
 
 	return nil
 }
 
+// CloseChns closes update channels
+func (t *Textile) CloseChns() {
```
sanderpick: and then we need to be able to shutdown after stopping

sanderpick: (only useful from the daemon)

textile.go
```diff
@@ -369,6 +369,8 @@ func buildNode(pinCode string, repoPath string) error {
 }
 
 func startNode() error {
+	listener := node.GetThreadUpdateListener()
```
sanderpick: because the channels are already created, we can create the listener here

textile.go
```diff
@@ -416,7 +417,6 @@ func startNode() error {
 						cmd.Green(desc) + cmd.Grey(" update to thread "+thrd)
 					fmt.Println(msg)
 				}
-			default:
```
sanderpick: will crush your fan!

mobile/mobile.go
```diff
@@ -139,6 +141,7 @@ func NewTextile(config *RunConfig, messenger Messenger) (*Mobile, error) {
 		RepoPath:  config.RepoPath,
 		node:      node,
 		messenger: messenger,
+		listener:  node.GetThreadUpdateListener(),
```
sanderpick: can live before the node is started and after its stopped, but we need a handle on it, so putting here

textile.go
carsonfarmer: nice catch 👍. that was there because otherwise it'll block until an update comes... but with a buffer of 10 like you added, this should probably be ok now right?

sanderpick: For me anyway, as simple as channels seem, they are weirdly unintuitive. In this case, we want the select to block. There's nothing else to do in the go routine besides wait for values. The default case is useful when reading / writing _a single_ value off an un-buffered channel that may not have a corresponding reader / writer, and is therefore blocking... so the default lets you sorta skip over it. Here though, the default is just causing the program to re-do the for loop in an unbounded manner, maxing cpu. 

As for the _the channel itself blocking_... A channel blocks _writes_ when it's not buffered or when it's buffer is full. So, as long as there's a listener, i.e., something pulling values out of the channel (like we're doing here) writes won't be blocked. The buffer ensure we don't _have_ to have a corresponding receiver for up to buffer length values... not really needed here because we're always gonna be receiving, but seems like best practice. In go by example's words: "By default channels are unbuffered, meaning that they will only accept sends (chan <-) if there is a corresponding receive (<- chan) ready to receive the sent value. Buffered channels accept a limited number of values without a corresponding receiver for those values." Not accept means block...

carsonfarmer: Cool, this all seems reasonable, thanks for the explanation 👍 _opens 'go by example' and gets reading_

PR #384, opened 2018-12-05
Adds messages and chat command

sanderpick:
- `messages` is like `comments` without a target, added to top level of thread
- leverages `messages` and `sub` commands into `chat` 💯

### DIFF
--- a/cmd/chat.go
+++ b/cmd/chat.go
@@ -1,0 +1,118 @@
+package cmd
+
+import (
+	"strings"
+
+	"github.com/chzyer/readline"
+)
+
+func init() {
+	register(&chatCmd{})
+}
+
+type chatCmd struct {
+	Client ClientOptions `group:"Client Options"`
+	Thread string        `short:"t" long:"thread" description:"Thread ID. Omit for all."`
+}
+
+func (x *chatCmd) Name() string {
+	return "chat"
+}
+
+func (x *chatCmd) Short() string {
+	return "Start a thread chat"
+}
+
+func (x *chatCmd) Long() string {
+	return `
+Starts an interactive chat session in a thread.
+Omit the --thread option to use the default thread (if selected).
+`
+}
+
+func (x *chatCmd) Execute(args []string) error {
+	setApi(x.Client)
+
+	if x.Thread == "" {
+		x.Thread = "default"
+	}
+
+	pid, username, err := getUsername()
+	if err != nil {
+		return err
+	}
+
+	rl, err := readline.New(Green(username + "  "))
+	if err != nil {
+		panic(err)
+	}
+	defer rl.Close()
+
+	updates, err := callSub(x.Thread, []string{"message"})
+	if err != nil {
+		return err
+	}
+
+	last := true
+	go func() {
+		for {
+			select {
+			case update, ok := <-updates:
+				if !ok {
+					return
+				}
+				if update.Block.AuthorId != pid {
+					if last {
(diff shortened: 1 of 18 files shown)


sanderpick:
Couldn't resist. Took a break from deploying. Will pick it up tomorrow.

sanderpick:
Check it out... threads chat, recovering from a member going offline. This feels like the culmination of the threads and cafe protocols.

screenshot 2018-12-04 20 51 14

cmd/sub.go
```diff
@@ -16,68 +15,111 @@ func init() {
 type subCmd struct {
 	Client ClientOptions `group:"Client Options"`
 	Thread string        `short:"t" long:"thread" description:"Thread ID. Omit for all."`
-	Type   string        `short:"k" long:"type" description:"Comma-separated list of event types to include. Omit for all."`
+	Type   []string      `short:"k" long:"type" description:"An update type to filter for. Omit for all."`
```
carsonfarmer: doh, didn't realize this was possible. this is what i wanted all along! nice 👍

cmd/sub.go
```diff
@@ -16,68 +15,111 @@ func init() {
...
+			output(string(data))
+		}
 	}
-	defer req.Body.Close()
-	if req.StatusCode >= 400 {
-		res, err := unmarshalString(req.Body)
+	return nil
+}
+
+func callSub(threadId string, types []string) (<-chan core.ThreadUpdate, error) {
```
carsonfarmer: yeh this is nice, glad its proving useful outside of demos!

cmd/sub.go
```diff
@@ -16,68 +15,111 @@ func init() {
...
-	}
-	decoder := json.NewDecoder(req.Body)
-	for {
-		var info core.ThreadUpdate
-		if err := decoder.Decode(&info); err == io.EOF {
-			break
-		} else if err != nil {
-			return err
+
+		if req.StatusCode >= 400 {
```
carsonfarmer: ah good catch, this came up today during some tests actually 👍

core/api.go
```diff
@@ -132,6 +133,12 @@ func (a *api) Start() {
 			}
 		}
 
+		messages := v0.Group("/messages")
```
carsonfarmer: this feels a bit weird having the two separate paths like this? not sure how else it should be done... is this pretty standard practice?

core/api_messages.go
```diff
@@ -0,0 +1,99 @@
+package core
+
+import (
+	"net/http"
+	"strconv"
+
+	"github.com/gin-gonic/gin"
+)
+
+func (a *api) addThreadMessages(g *gin.Context) {
```
carsonfarmer: so good!

cmd/chat.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (x *chatCmd) Long() string {
+	return `
+Starts an interactive chat session in a thread.
+Omit the --thread option to use the default thread (if selected).
+`
+}
+
+func (x *chatCmd) Execute(args []string) error {
```
carsonfarmer: siiiick!

core/api.go
sanderpick: yeah, agreed. same for files. my thought there was it's nice to be able to paginate across all files / messages... but maybe not needed. can revisit.

sanderpick:
Cheers for the look @carsonfarmer !

PR #387, opened 2018-12-05
cors(feature): adds cors support via textile config

carsonfarmer:
Take a look at this one @sanderpick. Should be pretty straight forward. Should be reasonable defaults. Basically just adds relevant CORS headers as Gin middle-wear. By default, leaves Allowed-Origin as length-one slice with API address, and Methods and Headers response Headers contain what you'd expect. Easy to change defaults, tried to keep these _out_ of the default config setup though.

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -349,7 +349,7 @@ func (a *api) abort500(g *gin.Context, err error) {
 	g.String(http.StatusInternalServerError, err.Error())
 }
 
-// getCORSSettings returns custom CORS settings given HTTPHeaders options
+// getCORSSettings returns custom CORS settings given HTTPHeaders config options
 func getCORSSettings(config *config.Config) cors.Config {
 	headers := config.API.HTTPHeaders
 	cconfig := cors.DefaultConfig()
@@ -361,10 +361,10 @@ func getCORSSettings(config *config.Config) cors.Config {
 			if origin == "*" {
 				cconfig.AllowAllOrigins = true
 				cconfig.AllowOrigins = nil
+				break
 			}
 		}
 	} else {
-		// By default use API origin host
 		defaultHost := config.Addresses.API
 		match, _ := regexp.MatchString("^https?://", defaultHost)
 		if !match {
@@ -376,17 +376,11 @@ func getCORSSettings(config *config.Config) cors.Config {
 	control, ok = headers["Access-Control-Allow-Methods"]
 	if ok && len(control) > 0 {
 		cconfig.AllowMethods = control
-	} else {
-		cconfig.AllowMethods = []string{"GET", "POST", "DELETE", "OPTIONS"}
 	}
 
 	control, ok = headers["Access-Control-Allow-Headers"]
 	if ok && len(control) > 0 {
 		cconfig.AllowHeaders = control
-	} else {
-		// default already has "Origin", "Content-Length", "Content-Type"
-		textileHeaders := []string{"Method", "X-Textile-Args", "X-Textile-Opts", "X-Requested-With", "Access-Control-Allow-Headers"}
-		cconfig.AllowHeaders = append(cconfig.AllowHeaders, textileHeaders...)
 	}
 
 	return cconfig
--- a/core/api.go
+++ b/core/api.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"mime/multipart"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


core/api.go
```diff
@@ -344,3 +341,46 @@ func (a *api) getFileConfig(g *gin.Context, mill m.Mill, use string, plaintext b
...
 }
+
+// getCORSSettings returns custom CORS settings given HTTPHeaders options
+func getCORSSettings(config *config.Config) cors.Config {
+	headers := config.API.HTTPHeaders
+	cconfig := cors.DefaultConfig()
+
+	control, ok := headers["Access-Control-Allow-Origin"]
+	if ok && len(control) > 0 {
+		cconfig.AllowOrigins = control
```
carsonfarmer: See https://github.com/gin-contrib/cors/issues/37

core/api.go
```diff
@@ -344,3 +341,46 @@ func (a *api) getFileConfig(g *gin.Context, mill m.Mill, use string, plaintext b
...
+			defaultHost = "http://" + defaultHost
+		}
+		cconfig.AllowOrigins = []string{defaultHost}
+	}
+
+	control, ok = headers["Access-Control-Allow-Methods"]
+	if ok && len(control) > 0 {
+		cconfig.AllowMethods = control
+	} else {
+		cconfig.AllowMethods = []string{"GET", "POST", "DELETE", "OPTIONS"}
```
carsonfarmer: Should be good set of defaults?

core/api.go
```diff
@@ -344,3 +341,46 @@ func (a *api) getFileConfig(g *gin.Context, mill m.Mill, use string, plaintext b
...
+	}
+
+	control, ok = headers["Access-Control-Allow-Headers"]
+	if ok && len(control) > 0 {
+		cconfig.AllowHeaders = control
+	} else {
+		// default already has "Origin", "Content-Length", "Content-Type"
+		textileHeaders := []string{"Method", "X-Textile-Args", "X-Textile-Opts", "X-Requested-With", "Access-Control-Allow-Headers"}
```
carsonfarmer: Seems like a reasonable set? I think this is all we ever need to deal with?

core/api.go
```diff
@@ -344,3 +341,46 @@ func (a *api) getFileConfig(g *gin.Context, mill m.Mill, use string, plaintext b
...
+		for _, origin := range control {
+			if origin == "*" {
+				cconfig.AllowAllOrigins = true
+				cconfig.AllowOrigins = nil
+			}
+		}
+	} else {
+		// By default use API origin host
+		defaultHost := config.Addresses.API
+		match, _ := regexp.MatchString("^https?://", defaultHost)
```
carsonfarmer: This feels messy to me, but the cors package requires http{s} prefixes on origins... ipfs for example, hard-codes:
localhosts: https://github.com/ipfs/go-ipfs/blob/af73c5097ce0ff70a18db25f801b34fa66a98202/core/corehttp/commands.go#L40

sanderpick:
Nice! Just took a quick look... I was imagining adding something to the textile config as defaults? Just so that people can see that as a tweakable setting when looking at the config... IPFS has these defaults for the gateway:

```
"Gateway": {
    "APICommands": [],
    "HTTPHeaders": {
      "Access-Control-Allow-Headers": [
        "X-Requested-With",
        "Range"
      ],
      "Access-Control-Allow-Methods": [
        "GET"
      ],
      "Access-Control-Allow-Origin": [
        "*"
      ]
    },
    "PathPrefixes": [],
    "RootRedirect": "",
    "Writable": false
  },
```

carsonfarmer:
Oh yeh for sure can add defaults to config... was thinking we'd want to keep these out, but that actually makes things simpler, so sure 👍. Also, I haven't touched the Gateway stuff yet, but can also add that to this branch.

sanderpick:
Ok nice. I'd say skip the gateway. I was just using that as an example snippet from the config... IPFS does't put those as defaults for the API... dunno why.

carsonfarmer:
probs because you don't want anything other than localhost to access your API 99% of the time?

repo/config/init.go
```diff
@@ -101,6 +101,13 @@ func Init(version string) (*Config, error) {
 		API: API{
 			HTTPHeaders: map[string][]string{
 				"Server": {"textile-go/" + version},
+				"Access-Control-Allow-Methods": []string{
+					"GET", "POST", "DELETE", "OPTIONS",
+				},
+				"Access-Control-Allow-Headers": []string{
+					"Content-Type", "Method", "X-Textile-Args", "X-Textile-Opts", "X-Requested-With",
```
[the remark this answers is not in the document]
sanderpick: boom

PR #389, opened 2018-12-06
Checks for local Username when returning ContactUsername

### LINKED ISSUE #388: Own files, commets, likes data with incorrect username
The `username` property of `ThreadFilesInfo`, `ThreadCommentInfo`, and `ThreadLikesInfo` is coming back as the first part of the peer id for items that are your own. Other peer's data looks correct.

carsonfarmer:
Fixes #388.

### DIFF
--- a/core/contacts.go
+++ b/core/contacts.go
@@ -79,6 +79,10 @@ func (t *Textile) Contacts() ([]ContactInfo, error) {
 // ContactUsername returns the username for the peer id if known
 func (t *Textile) ContactUsername(id string) string {
 	if id == t.node.Identity.Pretty() {
+		username, err := t.Username()
+		if err == nil && *username != "" {
+			return *username
+		}
 		return ipfs.ShortenID(id)
 	}
 	return toUsername(t.datastore.Contacts().Get(id))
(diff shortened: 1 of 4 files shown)


core/contacts.go
```diff
@@ -79,6 +79,10 @@ func (t *Textile) Contacts() ([]ContactInfo, error) {
 // ContactUsername returns the username for the peer id if known
 func (t *Textile) ContactUsername(id string) string {
 	if id == t.node.Identity.Pretty() {
+		username, err := t.Username()
+		if err == nil && *username != "" {
```
carsonfarmer: `t.Username` returns `*string` here, so is this right?

carsonfarmer:
Here ya go @asutula!

carsonfarmer:
oops!

sanderpick:
oh Carson, you should bump version in core/main.go, package.json, and mobile/package.json

asutula:
So this contact username will find it's way into all the different types we return from the api that include a `username` property?

carsonfarmer:
Yes, this is the only place that (for example) `ThreadFilesInfo`, `ThreadCommentInfo`, and `ThreadLikesInfo` ever get this information...

asutula:
Peeeeerfect. Thanks!

PR #390, opened 2018-12-06
docker: adds docker file

sanderpick:
Docker file is not done, but pushed this up cause it sneak in a fix for not being able to discover a public IP on an ec2 instance.


sanderpick:
(and I need a good linux/amd64 build to test)

PR #393, opened 2018-12-08
inbox improvements, misc.

sanderpick:
- Adds cafe message download attempt handling, currently set to 5 (they are deleted after 5 attempts)
- Improved message batch handling (not blocking when one bad message is repeated requested from the network).
- Adds cafe "you have mail" message used by hosts to ping clients they have a message. Because the host are servers (always online), there's a greater likelihood they have a direct connect with the intended client, more so even since they are bootstraps as well.
- Switches back to old IPNS settings and "floodsub" routing, which seems to really help publishing (we had been on the new and experimental "gossippub".
- Fix for the misleading "thread not found" error.
- Tweaks `wallet init` output allowing a node to be auto inited w/ the first account via a pipe: `textile init -s $(textile wallet init | tail -n1)` (w/o this, initing in docker is impossible cause you need to c/p the seed)


sanderpick:
Chat between two peers that I temp. forced to _always_ send to each others' inboxes (vs. direct). Pretty snappy with the "you have mail" message in the mix.

screenshot 2018-12-08 10 35 37

Issue #381, opened 2018-12-04
Logs command


sanderpick:
We need a command like IPFS for tweaking subsystem log level (take a look at their cmd interface)

carsonfarmer:
Fixed w/ #398.

PR #398, opened 2018-12-11
Adds verbosity controls to Textile logs

carsonfarmer:
Adds verbosity controls to Textile logs for running daemon, as well as when starting a new daemon.

Here is the running daemon command:

```
$ textile logs --help
Usage:
  textile [OPTIONS] logs [logs-OPTIONS]

List or change the verbosity of one or all subsystems log output.

Use the --subsystem option to control a given subsystem's log level. Omit to
list/change all subsystems.
Use the --level option to control the log level. One of: debug, info, warning, error,
critical. Omit to get current level.
Use the --tex-only flag to list/edit Textile subsystems only, otherwise, include all
available subsystems.

Textile logs piggyback on the IPFS event logs, so this command allows users to control
specific subsystem logs for finer control.

Help Options:
  -h, --help             Show this help message

[logs command options]
      -s, --subsystem=   The subsystem logging identifier. Omit for all.
      -l, --level=       The log level, with 'debug' the most verbose and 'critical'
                         the least verbose.
      -t, --tex-only     Whether to list/change only Textile subsystems, or all
                         available subsystems.

    Client Options:
          --api=         API address to use (default: http://127.0.0.1:40600)
          --api-version= API version to use (default: v0)
```

Also added controls to daemon subcommand:

```
$ textile daemon --help
Usage:
  textile [OPTIONS] daemon [daemon-OPTIONS]

Start a node daemon session.

Help Options:
  -h, --help          Show this help message

[daemon command options]
      -p, --pin-code= Specify the pin code for datastore encryption (omit of none was
                      used during init).
      -r, --repo-dir= Specify a custom repository path.
      -l, --logs=     Control subcommand log level. e.g., --logs="tex-core: debug" Can
                      be used multiple times.
```

Here's an example of using `logs` when starting the daemon. Note, the last `logs` flag will be silently ignored because it is a) not a valid subsystem, and b) not valid syntax, whereas the first two are both using valid syntax:

```
$ textile daemon --logs tex-core:debug --logs "tex-ipfs: info" --logs blah: nothing
```

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -1,0 +1,61 @@
+package cmd
+
+import "strconv"
+
+func init() {
+	register(&logsCmd{})
+}
+
+type logsCmd struct {
+	Client    ClientOptions `group:"Client Options"`
+	Subsystem string        `short:"s" long:"subsystem" description:"The subsystem logging identifier. Omit for all."`
+	Level     string        `short:"l" long:"level" description:"The log level, with 'debug' the most verbose and 'critical' the least verbose."`
+	All       bool          `short:"a" long:"all" description:"Whether to list/change only Textile subsystems (default: false), or all available subsystems (true)."`
+}
+
+func (x *logsCmd) Name() string {
+	return "logs"
+}
+
+func (x *logsCmd) Short() string {
+	return "List and control Textile subsystem logs."
+}
+
+func (x *logsCmd) Long() string {
+	return `List or change the verbosity of one or all subsystems log output.
+
+Use the --subsystem option to control a given subsystem's log level. Omit to list/change all subsystems.
+Use the --level option to control the log level. One of: debug, info, warning, error, critical. Omit to get current level.
+Use the --all flag to include all available subsystems. Omit to list/edit Textile subsystems only.
+
+Textile logs piggyback on the IPFS event logs, so this command allows users to control specific subsystem logs for finer control.
+`
+}
+
+func (x *logsCmd) Execute(args []string) error {
+	setApi(x.Client)
+	opts := map[string]string{
+		"subsystem": x.Subsystem,
+		"level":     x.Level,
+		"all":       strconv.FormatBool(x.All),
+	}
+	return callLogs(opts)
+}
+
+func callLogs(opts map[string]string) error {
+	subsystem := opts["subsystem"]
+	if subsystem != "" {
+		subsystem = "/" + subsystem
+	}
+	method := GET
+	if opts["level"] != "" {
+		method = POST
+	}
+	var info []map[string]string
+	res, err := executeJsonCmd(method, "logs"+subsystem, params{opts: opts}, &info)
+	if err != nil {
+		return err
+	}
+	output(res)
+	return nil
+}
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -10,7 +10,7 @@ type logsCmd struct {
 	Client    ClientOptions `group:"Client Options"`
 	Subsystem string        `short:"s" long:"subsystem" description:"The subsystem logging identifier. Omit for all."`
 	Level     string        `short:"l" long:"level" description:"The log level, with 'debug' the most verbose and 'critical' the least verbose."`
-	All       bool          `short:"a" long:"all" description:"Whether to list/change only Textile subsystems (default: false), or all available subsystems (true)."`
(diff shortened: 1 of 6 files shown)


GitCop:

 * Commit: f7a8874623a64b0943fb8e05e32d84c52df28769
  * Your subject line is longer than 80 characters

 * Commit: e957ab523794085a99d65e44279b045d471630ff
  * Your subject line is longer than 80 characters



---

repo/config/init.go
```diff
@@ -120,7 +120,6 @@ func Init(version string) (*Config, error) {
 		},
 		Logs: Logs{
 			LogToDisk: true,
-			LogLevel:  "error",
```
carsonfarmer: shouldn't need to worry about existing configs by removing this i think?

textile.go
```diff
@@ -274,7 +275,15 @@ func (x *migrateCmd) Execute(args []string) error {
 }
 
 func (x *daemonCmd) Execute(args []string) error {
-	if err := buildNode(x.PinCode, x.RepoPath); err != nil {
+	logLevels := make(map[string]string)
+	for _, log := range x.Logs {
+		split := strings.SplitN(log, ":", 2)
```
carsonfarmer: if the syntax isn't right, it'll be silently ignored... this seems appropriate? could be convinced to add validation..?

cmd/logs.go
```diff
@@ -0,0 +1,61 @@
+package cmd
+
+import "strconv"
+
+func init() {
+	register(&logsCmd{})
+}
+
+type logsCmd struct {
```
carsonfarmer: this one command is sort of a utility knife command. depending on the args, it does everything we need with logs...

core/api_logs.go
```diff
@@ -0,0 +1,54 @@
...
+			// validate subsystem + set log level
+			err = logging.SetLogLevel(system, llevel.String())
+			if err != nil {
+				g.String(http.StatusBadRequest, err.Error())
+				return
+			}
+			result = append(result, map[string]string{system: llevel.String()})
+		}
+	}
+	g.JSON(http.StatusOK, result)
```
carsonfarmer: this returns an array of `{subsystem: level}` pairs... would this be more useful as `{name: subsystem, level: level}` pairs?

carsonfarmer:
fixes #381

core/api_logs.go
sanderpick: seems good to me. what does "name" represent?

repo/config/init.go
sanderpick: perfect

textile.go
sanderpick: feels right, yeah

core/api_logs.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		a.abort500(g, err)
+		return
+	}
+	level := strings.ToUpper(opts["level"])
+	all := opts["all"] == "true"
+
+	var result []map[string]string
+	for _, system := range subsystems {
+		if strings.HasPrefix(system, "tex") || all {
```
sanderpick: would be nice to be able to drill into IPFS systems as well... as in, info _only_ from pubsub or something. does this restrict that?

core/api_logs.go
carsonfarmer: name would just be the key, as would level, so it would actually be something like:
```
[
  { "name": subsystem, "level", level },
  ...
]
```

But I think we can stick with:

```
[
  { subsystem: level },
  ...
]
```

sanderpick: ah right. yeah, feels right.

core/api_logs.go
carsonfarmer: nope, you can still do something like:

```
textile logs -s pusub -l info --all
```

to change only `"pubsub"` to `"info"` (note the `--all` flag). Reading that now though, maybe `all` isn't the right name for it?

carsonfarmer: this is even more straight forward with latest push

carsonfarmer:
api response in a bit more compact, returning a JSON object keyed by subsystem, rather than a JSON array with {key: value} pairs. should be easier to use, and much less verbose on cli

carsonfarmer:
Should be GTG @sanderpick. Just needs a final check I guess?

PR #405, opened 2018-12-13
URL encodes all cli args and opts

### LINKED ISSUE #381: Logs command

carsonfarmer:
simple fix. Fixes #381

### DIFF
--- a/cmd/chat.go
+++ b/cmd/chat.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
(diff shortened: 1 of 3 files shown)


carsonfarmer:
This will have to be documented in our API docs once we have those... but is a reasonable requirement (that all `X-Textile-*` args/opts are url escaped).

PR #408, opened 2018-12-14
Adds ContactUsername to thread, which doesn't have access to parent Textile instance

carsonfarmer:
Since a `Thread` doesn't know about the parent `Textile` instance, the `ContactUsername` method on `Textile` isn't accessible here. This copies that code, and adds it to `Thread`. I would be much nicer to have this in one place, hence why I'm not all that happy with this copy-pasta version, but it works and is simple.

### DIFF
--- a/core/thread.go
+++ b/core/thread.go
@@ -166,20 +166,11 @@ func (t *Thread) Info() (*ThreadInfo, error) {
 	if mod.Head != "" {
 		h := t.datastore.Blocks().Get(mod.Head)
 		if h != nil {
-			var username string
-			if len(h.AuthorId) >= 7 {
-				username = h.AuthorId[len(h.AuthorId)-7:]
-			}
-			contact := t.datastore.Contacts().Get(h.AuthorId)
-			if contact != nil && contact.Username != "" {
-				username = contact.Username
-			}
-
 			head = &BlockInfo{
 				Id:       h.Id,
 				ThreadId: h.ThreadId,
 				AuthorId: h.AuthorId,
-				Username: username,
+				Username: t.contactUsername(h.AuthorId),
 				Type:     h.Type.Description(),
 				Date:     h.Date,
 				Parents:  h.Parents,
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 7fb7e1c86d85f6eacabc172124ef9a15aa77c03c
  * Your subject line is longer than 80 characters



---

carsonfarmer:
Open to alternatives or better ideas entirely.

carsonfarmer:
Of course, _this_ could be the one version, but it seems very restrictive to require a `Thread` instance before being able to query for contact ids. The tricky part is, since we aren't one of our own contacts, we have to check _our_ `Identity` and test for that before checking for a contact (which would return `nil` for us). We _could_ add ourselves to our own contact store?

carsonfarmer:
Could also just use a static method and pass in an IpfsNode and Datastore, _buuut_ this doesn't currently work for `Mobile`, which doesn't have access to `m.node.node` or `m.node.datastore` because they aren't exported.

core/thread.go
```diff
@@ -648,3 +630,20 @@ func loadSchema(node *core.IpfsNode, id string) (*schema.Node, error) {
 	}
 	return &sch, nil
 }
+
+// contactUsername returns the username for the peer id if known
+// This is a near-exact port of the Textile version of this method
+func (t *Thread) contactUsername(id string) string {
```
sanderpick: We can still put this logic in one place `core/contacts.go`... just make it static and take the `datastore` and `peerId`.

sanderpick: (can take `self` too... our own peerId, and current own `username`)

carsonfarmer: yeh that might save it. taking those in a static method was something i tried before, but ran into issues on the mobile side because mobile doesn't have access to the unexported datastore variable... but maybe passing self and username will help me avoid that.

sanderpick:
Happy to just merge this and keep moving. Not the end of the world to have a couple lines of duplicated code.

PR #409, opened 2018-12-14
swarm: do not add ws by default

sanderpick:


sanderpick:
some small but important tweaks here.

PR #410, opened 2018-12-14
Switch to empty interface for JSON `Unmarshal`ing

carsonfarmer:
This allows us to support _any_ arbitrary top-level JSON primitives, including arrays. Useful for things like JSON Patch, which are arrays of JSON objects describing document modifications:

```
[
  {"op": "add", "path": "/question", "value": "????"}
  {"op": "add", "path": "/answer", "value": 42}
]
```


carsonfarmer:
I also removed the check for an empty JSON doc, because in some instances (like initializing a thread with an empty JSON object), we might want to allow this...

PR #412, opened 2018-12-14
Sander/outbox fix

sanderpick:


sanderpick:
cowboy style here, not much changing just setting the p2p wire limit to 0

PR #416, opened 2018-12-17
Adds peerid ndjson file to 005 migration

carsonfarmer:
hopefully this is useful 'as is' @sanderpick. adds tests as well. requires proper review due to the nature of the pr (for on-device migration).


sanderpick:
Looks good! Test failed though? Thanks for jumping in.

carsonfarmer:
Ah forgot to update test with improved JSON output. Should be GTG now @sanderpick  👍.

PR #422, opened 2018-12-18
[wip]: Adds contact cli api

carsonfarmer:
Initial proposal... if we like it, I can add `POST` endpoint for adding contacts...?

### DIFF
--- a/core/api_contacts.go
+++ b/core/api_contacts.go
@@ -1,0 +1,52 @@
+package core
+
+import (
+	"net/http"
+
+	"github.com/gin-gonic/gin"
+)
+
+func (a *api) lsContacts(g *gin.Context) {
+	opts, err := a.readOpts(g)
+	if err != nil {
+		a.abort500(g, err)
+		return
+	}
+
+	var contacts []ContactInfo
+
+	threadId := opts["thread"]
+	if threadId == "default" {
+		threadId = a.node.config.Threads.Defaults.ID
+	}
+	if threadId != "" {
+		thrd := a.node.Thread(threadId)
+		if thrd == nil {
+			g.String(http.StatusNotFound, ErrThreadNotFound.Error())
+			return
+		}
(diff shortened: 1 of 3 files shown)


carsonfarmer:
This should be GTG for a review now @sanderpick. **Note that these changes require a check for a propagated error on the mobile side of things.**

core/api_contacts.go
```diff
@@ -0,0 +1,90 @@
...
+	opts, err := a.readOpts(g)
+	if err != nil {
+		a.abort500(g, err)
+		return
+	}
+
+	id := args[0]
+	username := opts["username"]
+	if username == "" {
+		username = id[len(id)-7:]
```
sanderpick: wouldn't this be done _on the way out_? as in, when getting the contact, if username == "", use peerId

core/contacts.go
```diff
@@ -20,14 +20,16 @@ type ContactInfo struct {
...
 func (t *Textile) AddContact(id string, address string, username string) error {
-	if err := t.datastore.Contacts().Add(&repo.Contact{
+	err := t.datastore.Contacts().Add(&repo.Contact{
 		Id:       id,
 		Address:  address,
 		Username: username,
 		Added:    time.Now(),
-	}); err != nil {
+	})
+	if err != nil {
```
sanderpick: no need to log it here if we're returning it, could just return the datastore call... `return t.data....`

mobile/mobile_test.go
```diff
@@ -435,6 +435,13 @@ func TestMobile_AddContact(t *testing.T) {
 	}
 }
 
+func TestMobile_AddContactAgain(t *testing.T) {
```
sanderpick: nice!

core/contacts.go
carsonfarmer: yup

PR #431, opened 2018-12-23
Cafe public IP -> public address

sanderpick:
Since we're requiring the cafe to set a config variable containing it's publicly reachable IP, we may as well have it tag on http(s). This means the session's `HttpAddr` now contains `http(s)://`.

Note: I have not updated the demo cafes yet. Can do when mobile-side is ready.


asutula:
Cool, i'll update mobile as soon as a rc release is out.

PR #443, opened 2019-01-04
cafes: http registration

sanderpick:
Cafe registration over HTTP.


mobile/cafe.go
```diff
@@ -6,12 +6,12 @@ import (
 )
 
 // RegisterCafe calls core RegisterCafe
-func (m *Mobile) RegisterCafe(peerId string) error {
-	if !m.node.Online() {
-		return core.ErrOffline
+func (m *Mobile) RegisterCafe(host string) error {
+	if !m.node.Started() {
```
sanderpick: Since we're not relying on the swarm, the node just needs to be started to register, not online.

core/api_cafes.go
```diff
@@ -14,7 +14,7 @@ func (a *api) addCafes(g *gin.Context) {
 		return
 	}
 	if len(args) == 0 {
-		g.String(http.StatusBadRequest, "missing cafe id")
+		g.String(http.StatusBadRequest, "missing cafe host")
```
sanderpick: cli usage: `$ textile cafes add http://127.0.0.1:41601`

core/cafe_api.go
```diff
@@ -24,12 +24,14 @@ import (
...
 )
 
+// CafeInfo details info about this cafe
 type CafeInfo struct {
 	Peer     string `json:"peer"`
 	Address  string `json:"address"`
 	API      string `json:"api"`
 	Protocol string `json:"protocol"`
 	Node     string `json:"node"`
+	URL      string `json:"url"`
```
sanderpick: The URL needed for registration is provided here. Useful when doing the gateway cafe discovery thing.

mobile/cafe.go
asutula: cool, will be faster to launch the app the first time then

PR #457, opened 2019-01-09
protobuf CafeSession

asutula:
Still had to let the repo version of `CafeSession` continue to exist just so this change wouldn't creep too far into everything else.


sanderpick:
Nice, looks great. One nit is how to deal with the redundant named imports... I'm guessing VSCode wants those... I get a warning in GoLand:

screen shot 2019-01-09 at 1 46 37 pm

Buuuttt, I've seen PL repos say that you should always use named imports... not sure what the official opinion is there.

asutula:
Cool, yea let's talk about that and plenty of other things in here. This PR is def not g2g, just opened it as a place to discuss things and eventually merge.

core/threads.go
```diff
@@ -133,8 +133,8 @@ func (t *Textile) RemoveThread(id string) (mh.Multihash, error) {
 }
 
 // Threads lists loaded threads
-func (t *Textile) Threads() []Thread {
-	var threads []Thread
+func (t *Textile) Threads() []*Thread {
```
asutula: Made this unrelated looking change because of warning about copying a mutex or something

repo/db/cafe_sessions.go
```diff
@@ -37,11 +38,12 @@ func (c *CafeSessionDB) AddOrUpdate(session *repo.CafeSession) error {
 		return err
 	}
 
+	// TODO: There are props on pb.CafeSession we're not persisting
```
asutula: Ok here is one thing to make sure is right. There are props on `CafeSession` that aren't being persisted. Do we need to persist them all? If so, do we need to do anything with migration old session data?

repo/db/cafe_sessions.go
```diff
@@ -37,11 +38,12 @@ func (c *CafeSessionDB) AddOrUpdate(session *repo.CafeSession) error {
...
 	_, err = stmt.Exec(
 		session.Id,
 		session.Access,
 		session.Refresh,
-		session.Expiry.UnixNano(),
+		int(session.Exp.Nanos), // Nanos or Seconds?
```
asutula: I'm not clear about `Seconds` vs `Nanos` on a `Timestamp` object, and also know @sanderpick made some changes to get time related db things working correctly on Android. Want to highlight this and make sure we do the right think here.

mobile/cafe.go
```diff
@@ -18,48 +19,60 @@ func (m *Mobile) RegisterCafe(host string) error {
 }
 
 // CafeSessions calls core CafeSessions
-func (m *Mobile) CafeSessions() (string, error) {
+func (m *Mobile) CafeSessions() ([]byte, error) {
 	if !m.node.Started() {
-		return "", core.ErrStopped
+		return nil, core.ErrStopped
```
asutula: Tried this out, not being a go or c expert with pointers, but it seems to work. The fact that methods that return `error` can return `nil` when there is no error made me want to try returning `nil` instead of `make([]byte, 0)` in cases where there is no actual byte array to return.

mobile/cafe.go
```diff
@@ -18,48 +19,60 @@ func (m *Mobile) RegisterCafe(host string) error {
...
-	return toJSON(session)
+	return bytes, nil
 }
 
 // RefreshCafeSession calls core RefreshCafeSession
-func (m *Mobile) RefreshCafeSession(peerId string) (string, error) {
+func (m *Mobile) RefreshCafeSession(peerId string) ([]byte, error) {
 	if !m.node.Started() {
-		return "", core.ErrOffline
+		return nil, core.ErrStopped
```
asutula: Think we were using the wrong error return type here.

core/threads.go
sanderpick: very nice

mobile/cafe.go
sanderpick: Cool, makes sense to me. The "zero" value for a slice is `nil`

mobile/cafe.go
sanderpick: ah yeah, good eye

repo/db/cafe_sessions.go
sanderpick: I think we're okay as is. The repo session was originally meant to store just the pieces we needed in order to use it w/ requests and to easily check beforehand if it's expired to reduce round trips. In fact, we only _really_ need the two tokens themselves (access and refresh) because the other info can be parsed out of those, but that's a pita.

repo/db/cafe_sessions.go
sanderpick: Just to be consistent w/ the other persisted data, I'd say let's use nanos here even though we don't really need that resolution. Though, I think `Exp.Nanos` is just the nano second remainder (go's `UnixNano()` returns the entire delta from epoch as nanos) because it's only `int32`, `Exp.Seconds` is `int64`. We want `int64` nanos here. Maybe easiest to convert to a go `Time` then use `UnixNano()`.

repo/db/cafe_sessions.go
```diff
@@ -92,17 +94,22 @@ func (c *CafeSessionDB) handleQuery(stm string) []repo.CafeSession {
...
+		var rcafe *pb.Cafe
 		if err := json.Unmarshal(cafe, &rcafe); err != nil {
 			log.Errorf("error unmarshaling cafe: %s", err)
 			continue
 		}
 
-		ret = append(ret, repo.CafeSession{
+		timestamp := timestamp.Timestamp{
+			Seconds: expiryInt,
+			Nanos:   0,
```
sanderpick: And then `Nanos` can be used here.

asutula: I'm a little confused by this situation. I made the change where we're storing UnixNanos in the db... That means that `expiryInt` is nanoseconds from the epoch here. We're creating a pb `Timestamp`, where (my understanding is that) `Seconds` is seconds from epoch and `Nanos` are the remainder nanoseconds beyond that number of seconds. Is that true? If so, we need to do the math to get the number of seconds and remainder nanoseconds?

sanderpick: Right yeah, you'd have to do the math... but go's `time.Unix` method does that for you if you only supply nanos... so maybe easiest to just use that, then convert that to proto

asutula: I don't think that's true about `time.Unix`... The docs:
```Unix returns the local Time corresponding to the given Unix time, sec seconds and nsec nanoseconds since January 1, 1970 UTC. It is valid to pass nsec outside the range [0, 999999999].```

Comment (asutula) [reply repo/db/cafe_sessions.go]: Ahhh IS valid to pass nano outside that range. Read that wrong.

Comment (sanderpick) [reply repo/db/cafe_sessions.go]: Yeah, it's like, if nanos appear to _not be a remainer_, use them as such

File: repo/db/cafe_sessions.go [hunk 7]
```diff
@@ -41,7 +43,7 @@ func (c *CafeSessionDB) AddOrUpdate(session *repo.CafeSession) error {
 		session.Id,
 		session.Access,
 		session.Refresh,
-		session.Expiry.UnixNano(),
+		time.Unix(session.Exp.Seconds, int64(session.Exp.Nanos)).UnixNano(),
```
asutula: Ok, using `UnixNano()` here to store nanoseconds in the db.

repo/db/cafe_sessions.go
```diff
@@ -92,17 +94,24 @@ func (c *CafeSessionDB) handleQuery(stm string) []repo.CafeSession {
...
 			continue
 		}
 
-		ret = append(ret, repo.CafeSession{
+		time := time.Unix(0, expiryInt)
+
+		timestamp := timestamp.Timestamp{
+			Seconds: time.Unix(),
+			Nanos:   int32(time.Nanosecond()),
+		}
```
asutula: Pulling nanoseconds out of the db as `expiryInt`, creating a golang `Time` from it, then using that to create a proto `Timestamp`. Look right @sanderpick ?

asutula: Ah yea much cleaner, updated in both places.

PR #462, opened 2019-01-15
fix: ensure OS keeps nano precision in datastore

sanderpick:
Fixes int overflow on Android. Also updates all datastore date objects to use nanos, hopefully reducing confusion moving forward.

### DIFF
--- a/core/main.go
+++ b/core/main.go
@@ -34,7 +34,7 @@ import (
 var log = logging.Logger("tex-core")
 
 // Version is the core version identifier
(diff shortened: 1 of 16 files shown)


sanderpick:
In go, `int` always equals the processor bit-size... so, it's sorta like an alias to either `int32` or `int64`, though it's actually a concrete type. Surely, Android is not 32bit, but for some reason the cast was not resulting in `int` growing to `int64` like it was doing on other targets...

PR #464, opened 2019-01-16
contacts: fix bad joins

sanderpick:
Exposes an internal temp contact method that checks if the given peer ID matches the local nodes,
and if so, uses profile to build a contact. This is needed by a few cases where we're trying to grab the local node's contact info.


sanderpick:
Tested w/ two local peers and a local cafe.

PR #467, opened 2019-01-21
search: fix index out of bounds when deduping results

sanderpick:
Small little fix here... I noticed this when searching and getting no results. Since we don't have search yet in the app, no need to update the framework. This fix can just come along for the ride next time.


sanderpick:
Cheers!

PR #480, opened 2019-01-23
Adds contact directly to invite payload

sanderpick:
Also updates the logging system handling to allow debug during node init and creation, as well as it being controllable from the command line. This way, you can do stuff like:

```
$ textile init -s SXg67z8yjbocAhhg3Ct4tdaQLKVn2wxkt1K2G7v7UoHNEJED --debug -n -r ./boom
15:50:02.247  INFO   tex-repo: initializing repo at ./boom init.go:38
15:50:02.247  INFO   tex-ipfs: generating Ed25519 keypair for peer identity... crypto.go:14
Initialized account with address P8oq5vb8mMx93TKb5wzJcQP4VGnDG3WRNpmt4MxJbBkTSQDE
```


core/main.go
```diff
@@ -135,6 +138,14 @@ func InitRepo(conf InitConfig) error {
 		return ErrAccountRequired
 	}
 
+	logLevels := map[string]string{}
+	if conf.Debug {
+		logLevels = getTextileDebugLevels()
+	}
+	if _, err := setLogLevels(conf.RepoPath, logLevels, conf.LogToDisk); err != nil {
```
sanderpick: Unfortunately, I had to take this method off the node again because we need to use it before the `Textile` is created, here in init, but also in `NewTextile`, before it's finished being created (need to see debug output from migrations, for example)

core/main.go
```diff
@@ -466,6 +484,14 @@ func (t *Textile) LinksAtPath(path string) ([]*ipld.Link, error) {
 	return ipfs.LinksAtPath(t.node, path)
 }
 
+// SetLogLevels provides node scoped access to the logging system
+func (t *Textile) SetLogLevels(logLevels map[string]string) error {
```
sanderpick: However, i did add this method, which keep it exposed to mobile.

core/main.go
```diff
@@ -686,12 +712,21 @@ func (t *Textile) SetLogLevels(logLevels map[string]string) error {
...
+			return nil, err
 		}
 	}
-	t.writer = writer
-	return nil
+	return writer, nil
+}
+
+// getTextileDebugLevels returns a map of textile's logging subsystems set to debug
+func getTextileDebugLevels() map[string]string {
```
sanderpick: Useful with the `--debug` flag (which controls a bool on the `InitConfig`, and `RunConfig`)

core/thread_invites.go
```diff
@@ -68,11 +70,13 @@ func (t *Thread) AddExternalInvite() (mh.Multihash, []byte, error) {
 	if err != nil {
 		return nil, nil, err
 	}
+	contact := t.datastore.Contacts().Get(t.node().Identity.Pretty())
 	msg := &pb.ThreadInvite{
 		Sk:        threadSk,
 		Name:      t.Name,
 		Schema:    t.schemaId,
 		Initiator: t.initiator,
+		Contact:   repoContactToProto(contact),
```
sanderpick: This is needed because if you have never seen this contact before, you won't have their info. Also, we don't want to naively add it to the address book, so we just keep it on the indexed invite until it's been accepted.

repo/db/cafe_sessions.go
```diff
@@ -40,7 +39,7 @@ func (c *CafeSessionDB) AddOrUpdate(session *pb.CafeSession) error {
 		return err
 	}
 
-	time, err := ptypes.Timestamp(session.Exp)
+	exp, err := ptypes.Timestamp(session.Exp)
```
sanderpick: `time` conflicts w/ import name (goland warning)

repo/db/contacts_test.go
```diff
@@ -86,50 +93,83 @@ func TestContactDB_AddOrUpdate(t *testing.T) {
...
+	}
+}
+
+func TestContactDB_Count(t *testing.T) {
+	if contactStore.Count() != 2 {
+		t.Error("returned incorrect count of contacts")
+	}
+}
+
+func TestContactDB_UpdateUsername(t *testing.T) {
```
sanderpick: Some methods were missing from this test suite.

repo/init.go
```diff
@@ -25,7 +25,7 @@ var ErrRepoDoesNotExist = errors.New("repo does not exist, initialization is req
 var ErrMigrationRequired = errors.New("repo needs migration")
 var ErrRepoCorrupted = errors.New("repo is corrupted")
 
-const repover = "6"
```
sanderpick: eh, forgot to bump this last time. ndb cause the migration would just be applied on top

repo/migrations/migration007.go
```diff
@@ -0,0 +1,62 @@
...
+	"database/sql"
+	"os"
+	"path"
+
+	_ "github.com/mutecomm/go-sqlcipher"
+)
+
+type Minor007 struct{}
+
+func (Minor007) Up(repoPath string, pinCode string, testnet bool) error {
```
sanderpick: Drops `thread_invites` and re-adds the table with the new `contact` field.

textile.go
```diff
@@ -320,32 +331,6 @@ func getRepoPath(repoPath string) (string, error) {
 	return repoPath, nil
 }
 
-func buildNode(pinCode string, repoPath string, logLevels map[string]string) error {
```
sanderpick: No longer needed as an additional method.

Issue #479, opened 2019-01-23
Recursive message cli cmd calls list files after initial 'limit' is reached


carsonfarmer:
**Describe the bug**
The `callLsMessages` function has a recursive call in it to fetch more messages after `limit` has been reached. Currently, rather than recursively calling itself, it calls to `callLs`, which is the function to fetch thread files. This is a simple typo.

**To Reproduce**
1. `textile messages ls -l 1
2. Hit return

**Expected behavior**
Should continue to list messages, rather than files/error.

**Desktop (please complete the following information):**
 - OS: OSx
 - Textile Version 1.0.0-rc27

**Proposed fix**
Just fix typo and all will be well.

sanderpick:
oh! good eye

PR #481, opened 2019-01-24
Fix stop show list files when 'limit' is reached #479

u5surf:
#479 
I fix to stop showing file list when limit is reached.

### DIFF
--- a/cmd/messages.go
+++ b/cmd/messages.go
@@ -120,17 +120,21 @@ func callLsMessages(opts map[string]string) error {
 	if err != nil {
 		return err
 	}
-	if len(list) < limit {
+	if len(list) <= limit {
 		return nil
 	}
-
+	callLs(map[string]string{
+		"thread": opts["thread"],
+		"offset": list[len(list)].Id,
+		"limit":  opts["limit"],
+	})
 	reader := bufio.NewReader(os.Stdin)
 	fmt.Print("next page...")
 	if _, err := reader.ReadString('\n'); err != nil {
 		return err
 	}
 
-	return callLs(map[string]string{
+	return callLsMessages(map[string]string{
 		"thread": opts["thread"],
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: 2df95f4ec8366c435259779c12bd81729f54383c



---

cmd/messages.go
```diff
@@ -120,17 +120,21 @@ func callLsMessages(opts map[string]string) error {
...
+		"limit":  opts["limit"],
+	})
 	reader := bufio.NewReader(os.Stdin)
 	fmt.Print("next page...")
 	if _, err := reader.ReadString('\n'); err != nil {
 		return err
 	}
 
-	return callLs(map[string]string{
+	return callLsMessages(map[string]string{
```
sanderpick: I think this is the only change needed.

cmd/messages.go
```diff
@@ -120,17 +120,21 @@ func callLsMessages(opts map[string]string) error {
 	if err != nil {
 		return err
 	}
-	if len(list) < limit {
+	if len(list) <= limit {
 		return nil
 	}
-
+	callLs(map[string]string{
```
sanderpick: No need to call this here.

cmd/messages.go
```diff
@@ -120,17 +120,21 @@ func callLsMessages(opts map[string]string) error {
 	if err != nil {
 		return err
 	}
-	if len(list) < limit {
+	if len(list) <= limit {
```
sanderpick: If this is `<=`, you won't ever get anything past the first page. We could add some kind of "more" flag to the response to indicate that there are more results waiting on the node, but that would require some extra database work... may as well just make one more call and get an empty slice in the case where the total results equal limit.

sanderpick:
Cheers @u5surf !

PR #483, opened 2019-01-25
Respect thread types and add sharing style

sanderpick:
Adds handling for the currently coded but not handled thread types, and a new field `sharing`. The combination of access control + sharing control is a mind melt-er (my mind, at least). An additional "members" field restricts access to the given textile addresses across both dimensions. Please help me review the logic in code. From the updated thread command docs:

```
Thread type controls read (R), annotate (A), and write (W) access:

private   --> initiator: RAW, members:
read-only --> initiator: RAW, members: R
public    --> initiator: RAW, members: RA
open      --> initiator: RAW, members: RAW

Thread sharing style controls if (Y/N) a thread can be shared:

not-shared  --> initiator: N, members: N
invite-only --> initiator: Y, members: N
shared      --> initiator: Y, members: Y
```


sanderpick:
Anyone want to look at this?

asutula:
Going through that table above... seems right to me. Might be helpful to talk about a few use cases in Textile Photos:

Account thread: private, not-shared
Camera roll backup: private, not-shared
Open group (what's in the existing app): open, shared
Closed group (what we want to add): open, invite-only

Thinking it could be good documentation to create a table showing all the possible combinations of thread type and sharing style and a use case example for each. Some combinations are never valid (private + any share style), while others are valid and interesting under certain use cases (read-only + shared for a blog post with no commenting allowed).

One question... Does invite-only imply that the initiator can invite additional peers after the initial set of invites is sent out?

core/thread.go
```diff
@@ -608,6 +628,95 @@ func (t *Thread) pushUpdate(index BlockInfo) {
...
+// readable returns whether or not this thread is readable from the
+// perspective of the given address
+func (t *Thread) readable(addr string) bool {
+	if addr == t.initiator {
+		return true
+	}
+	switch t.ttype {
+	case repo.PrivateThread:
+		return false // should not happen
+	case repo.ReadOnlyThread:
```
carsonfarmer: nit: could use `fallthrough` here?

core/thread.go
```diff
@@ -608,6 +628,95 @@ func (t *Thread) pushUpdate(index BlockInfo) {
...
+// perspective of the given address
+func (t *Thread) annotatable(addr string) bool {
+	if addr == t.initiator {
+		return true
+	}
+	switch t.ttype {
+	case repo.PrivateThread:
+		return false // should not happen
+	case repo.ReadOnlyThread:
+		return false
```
carsonfarmer: same here

core/thread.go
```diff
@@ -608,6 +628,95 @@ func (t *Thread) pushUpdate(index BlockInfo) {
...
+	}
+}
+
+// shareable returns whether or not this thread is shareable from one address to another
+func (t *Thread) shareable(from string, to string) bool {
+	switch t.sharing {
+	case repo.NotSharedThread:
+		return false
+	case repo.InviteOnlyThread:
+		return from == t.initiator && t.member(to)
```
carsonfarmer: for this, can the initiator add members after the fact? i can't find that here? if no, then this seems quite limiting?

core/thread.go
```diff
@@ -608,6 +628,95 @@ func (t *Thread) pushUpdate(index BlockInfo) {
...
+		return false
+	}
+}
+
+// member returns whether or not the given address is a thread member
+// NOTE: Thread members are a fixed set of textile addresses specified
+// when a thread is created. If empty, _everyone_ is a member.
+func (t *Thread) member(address string) bool {
+	if len(t.members) == 0 {
+		return true
```
carsonfarmer: ah... ok, so membership is immutable or open. hmm, this might be too limiting?

core/thread_joins.go
```diff
@@ -79,6 +87,13 @@ func (t *Thread) handleJoinBlock(hash mh.Multihash, block *pb.ThreadBlock) (*pb.
 		return nil, err
 	}
 
+	if !t.readable(t.config.Account.Address) {
+		return nil, ErrNotReadable
+	}
+	if !t.readable(block.Header.Address) {
```
carsonfarmer: nice

sanderpick:
Ah yes, was thinking around this too. In the code here,

- Invite-only + a fixed member list = invites can only be sent to the member list by the initiator
- Invite-only + an empty member list = invites can be sent to anyone by the initiator

This feels right because _all_ the members should expect that the member list is either fixed or not. Imagine a 1-1 convo where the other person just invites a stranger in. If you had joined w/ the notion that the initiator _could_ invite others (not fixed member list) then you might expect that could happen and behave slightly differently... ?

core/thread.go
sanderpick: I thought that as well, but turns out you can't because what would you expect `t.ttype` to be when you arrive in another case......

core/thread.go
sanderpick: see comment below

core/thread.go
carsonfarmer: ah, ok

carsonfarmer:
Yup, totally makes sense @sanderpick, I'm convinced!

sanderpick:
But check me here... we _could_ allow the initiator to bump the members list, but wouldn't that just be the same as open totally open, because they're the only one that can invite anyway (in the invite-only case). I guess this is where the chart could be really handy. Probably some more overlapping combinations.

carsonfarmer:
Yeh, I think you're logic tracks there.

andrewxhill:
damn, so excited about this one! readonly ftw!

PR #490, opened 2019-01-26
Search and collect results sans Cafe

### LINKED ISSUE #489: Allowing searching and adding contacts sans Cafe
This will require:
- The client to directly search the network over pubsub
- Search results to flow in over pubsub w/ request ID as topic, vs. direct responses to the always online Cafe.

sanderpick:
fixes #489


cmd/contacts.go
```diff
@@ -116,12 +116,15 @@ type findContactsCmd struct {
...
 	Wait     int           `long:"wait" description:"Stops searching after 'wait' seconds have elapsed." default:"5"`
+	Add      bool          `long:"add" description:"Add results to local contacts. Not allowed when searching by username."`
 }
 
 func (x *findContactsCmd) Usage() string {
 	return `
 
-Finds contacts known locally and on the network.`
+Finds contacts known locally and on the network.
+Use the --add option to save remote results as local contacts.
```
sanderpick: add a contact from search results

core/cafe_service.go
```diff
@@ -1026,7 +1063,24 @@ func (h *CafeService) handlePubSubContactQuery(pid peer.ID, env *pb.Envelope) (*
 		return nil, nil
 	}
 
-	return h.service.NewEnvelope(pb.Message_CAFE_PUBSUB_CONTACT_QUERY_RES, res, nil, false)
+	renv, err := h.service.NewEnvelope(pb.Message_CAFE_PUBSUB_CONTACT_QUERY_RES, res, nil, false)
+	if err != nil {
+		return nil, err
+	}
+
+	switch query.ResponseType {
```
sanderpick: respond here w/ either direct p2p or pubsub, depending on what the caller wants / needs

sanderpick: (clients w/o cafes need pubsub because they don't have a public IP address)

core/thread.go
```diff
@@ -349,7 +349,7 @@ func (t *Thread) addOrUpdateContact(contact *repo.Contact) error {
 	}
 
 	ex := t.datastore.Contacts().Get(contact.Id)
-	if ex != nil && contact.Updated.UnixNano() >= ex.Updated.UnixNano() {
+	if ex == nil || ex.Updated.UnixNano() < contact.Updated.UnixNano() {
```
sanderpick: a bug! contacts would never be added from a join here, luckily never got released :)

service/main.go
```diff
@@ -506,3 +506,54 @@ func (srv *Service) listen() {
 		}
 	}
 }
+
+// ListenFor opens a tmp subscription to a topic and passes results to handler
+func (srv *Service) ListenFor(topic string, handler func(pid peer.ID, env *pb.Envelope) (*pb.Envelope, error)) context.CancelFunc {
```
sanderpick: Service layer provides this helper for spinning up a tmp subscription at a random ID (the search request ID in the case of contact search)

core/thread.go
carsonfarmer: nice find :)

PR #492, opened 2019-01-29
Add access token tooling for Cafe peers

carsonfarmer:
Adds tooling for creating, listing, validating, and removing access tokens.

- [x] Use tokens for 'cafes' POST API endpoint
- [x] Add 'tokens' API
- [x] Update Cafe service protobufs to include a token field
- [x] Require tokens for registration service


cmd/cafe.go
```diff
@@ -37,18 +37,25 @@ Use this command to add, list, get, remove cafes and check messages.
...
 	Client ClientOptions `group:"Client Options"`
+	Token  string        `required:"true" short:"t" long:"token" description:"An access token supplied by the Cafe."`
 }
 
 func (x *addCafesCmd) Usage() string {
 	return `
 
-Registers with a cafe and saves an expiring service session token.`
+Registers with a cafe and saves an expiring service session token.
+Unless explicitly disabled by the Cafe, an access token is required to register.
```
carsonfarmer: This isn't actually currently possible, but wanted to have it here for discussion. Any reason to allow a Cafe to disable tokens? Basically leave itself wide open for registrations?

cmd/cafe.go
```diff
@@ -37,18 +37,25 @@ Use this command to add, list, get, remove cafes and check messages.
 
 type addCafesCmd struct {
 	Client ClientOptions `group:"Client Options"`
+	Token  string        `required:"true" short:"t" long:"token" description:"An access token supplied by the Cafe."`
```
carsonfarmer: Left this as a required flag to make sure its explicit.

cmd/tokens.go
```diff
@@ -0,0 +1,124 @@
...
+var errMissingToken = errors.New("missing token")
+
+func init() {
+	register(&tokensCmd{})
+}
+
+type tokensCmd struct {
+	Create  createTokensCmd  `command:"create" description:"Create a new access token"`
+	List    lsTokensCmd      `command:"ls" description:"List available access tokens"`
+	Compare compareTokensCmd `command:"compare" description:"Check if access token is valid"`
```
carsonfarmer: Could be renamed to `check` and `checkTokensCmd`... or even `get` given that is what the underlying rest API uses?

cmd/tokens.go
```diff
@@ -0,0 +1,124 @@
...
+
+type createTokensCmd struct {
+	Client ClientOptions `group:"Client Options"`
+}
+
+func (x *createTokensCmd) Usage() string {
+	return `
+
+Generates an access token (32 random bytes) and saves bcrypt encrypted version for future lookup.
+The response contains a base58 encoded version of the random bytes token.
```
carsonfarmer: The token is actually a `ksuid() + "+" + base58(randomBytes)` string. Could be very explicit here, though code documents this fairly well.

cmd/tokens.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+type compareTokensCmd struct {
+	Client ClientOptions `group:"Client Options"`
+}
+
+func (x *compareTokensCmd) Usage() string {
+	return `
+
+Check validity of existing cafe developer access token.
```
carsonfarmer: Maybe should remove 'developer' nomenclature throughout? Doesn't _really_ mean much, as non-devs might also want to explicitly register with a given Cafe?

core/api.go
```diff
@@ -189,6 +189,14 @@ func (a *api) Start() {
 			cafes.POST("/messages", a.checkCafeMessages)
 		}
 
+		tokens := v0.Group("/tokens")
+		{
+			tokens.POST("", a.createTokens)
+			tokens.GET("", a.lsTokens)
+			tokens.GET("/:id", a.compareTokens)
```
carsonfarmer: Does this seems like a good RESTful approach here? Use get and delete on the token-based resource?

core/api_tokens.go
```diff
@@ -0,0 +1,47 @@
...
+func (a *api) createTokens(g *gin.Context) {
+	token, err := a.node.CreateCafeToken()
+	if err != nil {
+		a.abort500(g, err)
+		return
+	}
+	g.String(http.StatusCreated, token)
+}
+
+func (a *api) lsTokens(g *gin.Context) {
```
carsonfarmer: This one (and related cli tools etc) isn't really all that useful tbh. Since the Cafe never stores the actual token (only the bcrypt hashed version) listing them doesn't provide much utility beyond sanity checking the id portion...

core/cafe.go
```diff
@@ -9,8 +9,8 @@ import (
 
 // RegisterCafe registers this account with another peer (the "cafe"),
 // which provides a session token for the service
-func (t *Textile) RegisterCafe(host string) (*pb.CafeSession, error) {
-	session, err := t.cafe.Register(host)
+func (t *Textile) RegisterCafe(host string, token string) (*pb.CafeSession, error) {
```
carsonfarmer: token string now required

core/cafe_api_test.go
```diff
@@ -68,8 +68,21 @@ func TestCafeApi_Setup(t *testing.T) {
...
-	// register cafe
-	if _, err := node1.RegisterCafe("http://127.0.0.1:5000"); err != nil {
+	// create token on cafe
+	token, err := node2.CreateCafeToken()
+	if err != nil {
+		t.Error(fmt.Errorf("error creating cafe token: %s", err))
+		return
+	}
+
+	ok, err := node2.CompareCafeDevToken(token)
```
carsonfarmer: This is just an additional check in the test, not required to just get the test working.

core/cafe_service.go
```diff
@@ -235,7 +237,7 @@ func (h *CafeService) FindContactPubSub(query *pb.CafeContactQuery, fromCafe boo
 }
 
 // Register creates a session with a cafe
-func (h *CafeService) Register(host string) (*pb.CafeSession, error) {
+func (h *CafeService) Register(host string, token string) (*pb.CafeSession, error) {
```
carsonfarmer: Again, tokens required now

core/cafe_service.go
```diff
@@ -615,6 +618,28 @@ func (h *CafeService) handleRegistration(pid peer.ID, env *pb.Envelope) (*pb.Env
 		return h.service.NewError(403, errForbidden, env.Message.RequestId)
 	}
 
+	// does the provided token match?
+	s := strings.Split(reg.Token, "+")
```
carsonfarmer: tokens are `ksuid + "+" + base58(randomBytes)`

core/cafe_service.go
```diff
@@ -615,6 +618,28 @@ func (h *CafeService) handleRegistration(pid peer.ID, env *pb.Envelope) (*pb.Env
...
+	if len(s) != 2 {
+		h.service.NewError(403, errForbidden, env.Message.RequestId)
```
carsonfarmer: Could probably create a more nuanced error msg here? But this seemed sufficient?

core/cafe_api_test.go
```diff
@@ -68,8 +68,21 @@ func TestCafeApi_Setup(t *testing.T) {
...
+	if !ok || err != nil {
+		t.Error(fmt.Errorf("error checking token: %s", err))
+		return
+	}
+
+	// register with cafe
+	if _, err := node1.RegisterCafe("http://127.0.0.1:5000", token); err != nil {
```
sanderpick: sweet

core/cafe_service.go
```diff
@@ -615,6 +618,28 @@ func (h *CafeService) handleRegistration(pid peer.ID, env *pb.Envelope) (*pb.Env
...
+	if err != nil {
+		return h.service.NewError(403, errForbidden, env.Message.RequestId)
+	}
+
+	encodedToken := h.datastore.CafeDevTokens().Get(id)
+	if encodedToken == nil {
+		return h.service.NewError(403, errForbidden, env.Message.RequestId)
+	}
+
+	err = bcrypt.CompareHashAndPassword([]byte(encodedToken.Token), plainBytes)
```
carsonfarmer: We (for safety sake, must) use `CompareHashAndPassword`, which is why we use a concated token on the client side, and index them by their ksuid component. This makes it possible to pull out the token from the db, which contains the salt and hashing params. Then we can use the constant time comparison function from bcrypt to compare the tokens.

core/main_test.go
```diff
@@ -18,7 +19,24 @@ import (
 )
 
 var repoPath = "testdata/.textile"
+var otherPath = "testdata/.textile2"
 var node *Textile
+var other *Textile
+var token string
+var contact = &repo.Contact{
```
carsonfarmer: Pulled this stuff over from mobile as a 'bonus'

core/tokens.go
```diff
@@ -0,0 +1,91 @@
...
+// and stores a bcrypt hashed version for later comparison
+func (t *Textile) CreateCafeToken() (string, error) {
+	key, err := crypto.GenerateAESKey()
+	if err != nil {
+		return "", err
+	}
+
+	id := ksuid.New().String()
+	created := time.Now()
+	rawToken := key[:32]
```
carsonfarmer: We only need 32 bytes really, so just grab those. Easier than re-implementing the random reader function we already have.

core/tokens.go
```diff
@@ -0,0 +1,91 @@
...
+
+	safeToken, err := bcrypt.GenerateFromPassword(rawToken, bcrypt.DefaultCost)
+	if err != nil {
+		return "", err
+	}
+
+	err = t.datastore.CafeDevTokens().Add(
```
carsonfarmer: Only store bcrypt hashed version... db _never_ seems original random token.

core/tokens.go
```diff
@@ -0,0 +1,91 @@
...
+		&repo.CafeDevToken{
+			Id:      id,
+			Token:   safeToken,
+			Created: created,
+		})
+	if err != nil {
+		return "", err
+	}
+
+	return id + "+" + base58.FastBase58Encoding(rawToken), nil
```
sanderpick: what you're thinking w/ adding the id to the token?

core/tokens.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// CafeDevTokens lists all stored (bcrypt encrypted) dev tokens
+func (t *Textile) CafeDevTokens() ([]string, error) {
+	tokens := t.datastore.CafeDevTokens().List()
+	strings := make([]string, len(tokens))
+	for i, token := range tokens {
+		strings[i] = token.Id + "+" + base58.FastBase58Encoding(token.Token)
```
carsonfarmer: This is a bit silly maybe? Probably don't even need this method to be honest. Could be replaced with a `Count` method instead?

core/tokens.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	return strings, nil
+}
+
+// CompareCafeDevToken checks whether a supplied base58 encoded dev token matches the stored
+// bcrypt hashed equivalent
+func (t *Textile) CompareCafeDevToken(token string) (bool, error) {
```
carsonfarmer: Again, could be `Check*`

mobile/cafe.go
```diff
@@ -7,12 +7,12 @@ import (
 )
 
 // RegisterCafe calls core RegisterCafe
-func (m *Mobile) RegisterCafe(host string) error {
+func (m *Mobile) RegisterCafe(host string, token string) error {
```
carsonfarmer: Obviously will require modifications to the bridges. Also for our client libs, might be good to have default env-based tokens that we add?

repo/db/db.go
```diff
@@ -246,7 +253,9 @@ func initDatabaseTables(db *sql.DB, pin string) error {
 
     create table cafe_client_messages (id text not null, peerId text not null, clientId text not null, date integer not null, primary key (id, clientId));
     create index cafe_client_message_clientId on cafe_client_messages (clientId);
-    create index cafe_client_message_date on cafe_client_messages (date);
+	create index cafe_client_message_date on cafe_client_messages (date);
```
sanderpick: careful of the indenting here

repo/db/db.go
```diff
@@ -187,6 +193,7 @@ func initDatabaseTables(db *sql.DB, pin string) error {
 	if pin != "" {
 		sqlStmt = "PRAGMA key = '" + pin + "';"
 	}
+	// TODO: also add field to cafe_clients to link to cafe_dev_tokens
```
carsonfarmer: Haven't done this yet... but maybe we want it now to make migration easier?

repo/db/db.go
```diff
@@ -246,7 +253,9 @@ func initDatabaseTables(db *sql.DB, pin string) error {
 
     create table cafe_client_messages (id text not null, peerId text not null, clientId text not null, date integer not null, primary key (id, clientId));
     create index cafe_client_message_clientId on cafe_client_messages (clientId);
-    create index cafe_client_message_date on cafe_client_messages (date);
+	create index cafe_client_message_date on cafe_client_messages (date);
+		
+	create table cafe_dev_tokens (id text primary key not null, token text not null, created integer not null);
```
sanderpick: can we use `date` instead of `created`? i've been trying to stick to the convention of using date when there is only one, as in, no corresponding `updated` date.

sanderpick: you can add an index on token so that you can look it up directly

repo/migrations.go
```diff
@@ -76,7 +77,7 @@ func version(repoPath string) (int, error) {
 	} else if err != nil && os.IsNotExist(err) {
 		version = []byte("0")
 	}
-	v, err := strconv.Atoi(string(version[0]))
+	v, err := strconv.Atoi(string(version[:]))
```
carsonfarmer: This is potentially sig change. Since we're up to double-digits migrations, this was causing some unexpected migration issues (kept moving us back to 1).

sanderpick: nice

repo/migrations/migration009.go
```diff
@@ -0,0 +1,59 @@
...
+		return err
+	}
+	if pinCode != "" {
+		if _, err := db.Exec("pragma key='" + pinCode + "';"); err != nil {
+			return err
+		}
+	}
+
+	// add missing cafe_dev_token table
+	// TODO: should we also edit the cafe_client table to add a field linking to cafe_dev_token?
```
sanderpick: +1

cmd/cafe.go
sanderpick: sounds like an enhancement

cmd/tokens.go
sanderpick: how about `validate`?

core/tokens.go
sanderpick: seems nice to have to me

repo/db/db.go
sanderpick: i'd say this should be part of this PR, and put it into the same migration

repo/migrations.go
sanderpick: First time we've had double digit repover, so i doubt we've seen issues before.

core/cafe_service.go
sanderpick: I think we'll want to reconsider the concatenation - makes it feel overly complicated. You can look up the token by `token`, and an index to be more efficient

core/tokens.go
carsonfarmer: ok, let's keep it

cmd/cafe.go
carsonfarmer: Yup, agreed, removing text.

cmd/tokens.go
carsonfarmer: nice, changed

cmd/tokens.go
carsonfarmer: as per offline conversation, we'll change this to just use random bytes for id and avoid additional +

repo/migrations.go
carsonfarmer: I meant it was cause _me_ unexpected migration issues because I'm in the dds ;)

core/tokens.go
carsonfarmer: need it for db access on validation. as per our offline conversation, will update this to use random bytes instead of ksuid

repo/db/db.go
carsonfarmer: sure!

repo/db/db.go
carsonfarmer: yeh wth?

repo/migrations/migration009.go
carsonfarmer: Added

cmd/tokens.go
carsonfarmer: Yeh, I'm gonna do this.

core/cafe_service.go
carsonfarmer: changed to use random bytes for id and base58 encode the whole token

core/cafe_service.go
carsonfarmer: see other comments for why this isn't possible. however, have moved to different approach using random bytes for token and and base58 encoding the whole token

carsonfarmer:
Update and should be GTG @sanderpick:

```
$ textile tokens --help
Usage:
  textile [OPTIONS] tokens <command>

Tokens allow other peers to register with a Cafe peer.
Use this command to create, list, compare, and remove tokens required for access to
this peer's Cafe.


Help Options:
  -h, --help      Show this help message

Available commands:
  create    Create a new access token
  ls        List available access tokens
  rm        Remove a specific access token
  validate  Check if access token is valid
```
```
$ textile tokens create
CmKfx5pgcLgMRtMpR8r79KK88rH9GkndF7msQGKdoxmHhDfZ9q6cutnPoG5o
```
^ Hopefully those are pretty enough 😉

carsonfarmer:
For fun:
```
textile tokens create | xargs -I{} textile tokens validate {}
```

cmd/tokens.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+func (x *tokensCmd) Short() string {
+	return "Manage Cafe access tokens"
+}
+
+func (x *tokensCmd) Long() string {
+	return `
+Tokens allow other peers to register with a Cafe peer.
+Use this command to create, list, compare, and remove tokens required for access to this peer's Cafe.
```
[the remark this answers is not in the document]
carsonfarmer: fixed!

core/api_tokens.go
```diff
@@ -0,0 +1,52 @@
...
+		tokens = make([]string, 0)
+	}
+	g.JSON(http.StatusOK, tokens)
+}
+
+func (a *api) validateTokens(g *gin.Context) {
+	token := g.Param("id")
+	ok, err := a.node.ValidateCafeToken(token)
+	if err != nil || !ok {
+		g.String(http.StatusUnauthorized, "invlaid credentials")
```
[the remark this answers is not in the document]
carsonfarmer: fixed, good catch

core/tokens.go
```diff
@@ -0,0 +1,101 @@
...
+			return "", errors.New("invalid token format")
+		}
+	} else {
+		key, err = crypto.GenerateAESKey()
+		if err != nil {
+			return "", err
+		}
+	}
+	date := time.Now()
+	id := hex.EncodeToString(key[:12])
```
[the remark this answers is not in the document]
carsonfarmer: easier for comparison in db, and faster than base64/58/whatever. but no real good reason specifically beyond convenience

repo/migrations/migration009.go
```diff
@@ -0,0 +1,60 @@
...
```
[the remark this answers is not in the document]
carsonfarmer: removed

sanderpick:
Have you though about how we're gonna start using these in the mobile app? The app will have to know the code for every cafe. So, maybe a map like `<cafe_url, code>` could be added to the `.env` file? After cafe discovery hand over a primary and secondary, the app can just look up the right code from the local map. Something like that?

carsonfarmer:
Yup, we'll use the `.env` to specify a single, access code. This single code can be propagated to the various cafes via the new cli tools.

PR #493, opened 2019-01-29
ipfs: apply go-libp2p-kad-dht fix

sanderpick:
Includes go-ipfs w/ a patch from https://github.com/raulk/go-libp2p-kad-dht/pull/2 to go-libp2p-kad-dht@4.4.12.


sanderpick:
ok, finally go this to find the gx hashes in CI

PR #519, opened 2019-02-11
mobile: changes addThread api to include full config options

andrewxhill:


sanderpick:
Nice, looks great... simpler than I thought. Only nit is that it would be nice to use a proto object for the config because you'd get a TS object for free and the enums would be more descriptive, but we could always change later

sanderpick:
Eh, never mind... that would open can of worms. I think you'll need to not include the `repo` types in the config though because we'd have to then `gobind` that package as well. Probably easier to use the string versions of those types... you can convert back and forth, e.g.:

```
func (ts ThreadSharing) Description() string {
	switch ts {
	case NotSharedThread:
		return "NOT_SHARED"
	case InviteOnlyThread:
		return "INVITE_ONLY"
	case SharedThread:
		return "SHARED"
	default:
		return "INVALID"
	}
}

func ThreadSharingFromString(desc string) (ThreadSharing, error) {
	switch strings.ToUpper(strings.TrimSpace(desc)) {
	case "NOT_SHARED":
		return NotSharedThread, nil
	case "INVITE_ONLY":
		return InviteOnlyThread, nil
	case "SHARED":
		return SharedThread, nil
	default:
		return -1, errors.New("could not parse thread sharing")
	}
}
```

andrewxhill:
got it, was wondering best way there, was hoping go would just accept the typed strings or maybe just the Ints. will fiddle with that in a bit

andrewxhill:
Okay, take a look whenever

Issue #521, opened 2019-02-12
addThread schema doesn't support members whitelist


andrewxhill:
**Describe the bug**
go implementation uses a `string[]` which isn't supported in the build

message in the output

`skipped field MobileThreadConfig.Members with unsupported type: []string`

asutula:
We can go pbuf data or json string

Issue #520, opened 2019-02-12
Mobile framework is missing `searchContacts` method


andrewxhill:
**Describe the bug**
mobile framework is missing `searchContacts` method

Message in the build output

` skipped method Mobile.SearchContacts with unsupported parameter or return types`

sanderpick:
Darn, I forgot those objects need to be just `[]byte`

PR #523, opened 2019-02-12
Adds ApplyThreadBackup and some other cleanups

sanderpick:
Also fixes the mobile framework missing method issues. #520 and #521


mobile/contacts.go
```diff
@@ -70,8 +72,17 @@ func (m *Mobile) ContactThreads(id string) (string, error) {
 }
 
 // SearchContacts calls core SearchContacts
-func (m *Mobile) SearchContacts(query *pb.ContactQuery, options *pb.QueryOptions, cb Callback) (func(), error) {
-	resCh, errCh, cancel, err := m.node.SearchContacts(query, options)
+func (m *Mobile) SearchContacts(query []byte, options []byte, cb Callback) (*CancelFn, error) {
```
sanderpick: @andrewxhill I fixed this up here... turns out you can't just return `func()`, so I added this `CancelFn` object that will expose a single `call` method

mobile/search.go
```diff
@@ -6,17 +6,24 @@ import (
 	"github.com/textileio/textile-go/pb"
 )
 
+type CancelFn struct {
+	cancel *broadcast.Broadcaster
+	done   func()
+}
```
sanderpick: Here's this `CancelFn`... internally `call` wrangles the cancel broadcaster. Those params don't make it across the bridge, but `call` does, so all good

mobile/threads.go
```diff
@@ -20,16 +20,16 @@ type ExternalInvite struct {
...
-	Key      	string `json:"key"`
-	Name  	 	string `json:"name"`
-	Type  	 	string `json:"type"`
-	Sharing  	string `json:"sharing"`
-	Members   	[]string `json:"members"`
-	Schema	 	string `json:"schema"`
-	Media	 	bool `json:"media"`
-	CameraRoll	bool `json:"cameraRoll"`
+// AddThreadConfig is the mobile-client specific config for creating a new thread
+type AddThreadConfig struct {
```
sanderpick: `Mobile` gets prefixed by gomobile

mobile/threads.go
```diff
@@ -98,7 +96,7 @@ func (m *Mobile) AddThread(config *MobileThreadConfig) (string, error) {
 		Initiator: m.node.Account().Address(),
 		Type:      threadType,
 		Sharing:   sharingType,
-		Members:   config.Members,
+		Members:   util.SplitString(config.Members, ","),
```
sanderpick: just made members a comma separated string

mobile/contacts.go
```diff
@@ -70,8 +72,17 @@ func (m *Mobile) ContactThreads(id string) (string, error) {
 }
 
 // SearchContacts calls core SearchContacts
-func (m *Mobile) SearchContacts(query *pb.ContactQuery, options *pb.QueryOptions, cb Callback) (func(), error) {
```
sanderpick: I made these byte slices because you'll already have the types from re-generating the JS proto files. You can look at `addThreadFiles` for a ref, it does the same.

mobile/threads.go
andrewxhill: super

mobile/threads.go
andrewxhill: perfect

core/recovery.go
```diff
@@ -22,5 +31,105 @@ func (t *Textile) FindThreadBackups(query *pb.ThreadBackupQuery, options *pb.Que
...
+
+	if err := thrd.followParents([]string{backup.Head}); err != nil {
+		return err
+	}
+	hash, err := mh.FromB58String(backup.Head)
+	if err != nil {
+		return err
+	}
+
+	return thrd.updateHead(hash)
```
andrewxhill: super cool

PR #526, opened 2019-02-12
feed: adds annotated flag for nested/not-nested annotations

sanderpick:
Take a look at this @asutula - basically, adds a flag that dictated whether or not annotations (comments/likes) are nested underneath their targets or listed at the top level. If they're listed at the top level, their `target` field is populated w/ another feed item object which is the thing that comment/like is about.

### DIFF
--- a/cmd/feed.go
+++ b/cmd/feed.go
@@ -14,10 +14,11 @@ func init() {
 }
 
 type lsCmd struct {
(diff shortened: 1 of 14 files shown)


asutula:
Cool, will take a look asap

sanderpick:
Just pushed a test fix and made the flag available to the mobile framework as well.

PR #541, opened 2019-02-18
Account Sync APIs

### LINKED ISSUE #534: Account Sync
Here's a place where we can talk about the UX of account sync. Here's what I'm thinking to start:

### CLI
- Add a new `textile account` top-level command
- Add some more sub-commands:
  - `account address`: Moved from `textile address`, just prints account address
  - `account peers`: Lists all account peers
  - `account sync`: Searches for all thread backups by account address and starts loading HEAD from each
  - `account backups`: Moved from `threads backups`, searches for and prints all thread backups by account address
- Add a flag to `textile deamon` like `--sync` that performs a sync on start

### Mobile
- `AccountAddress`
- `AccountPeers`
- `AccountSync`
- `AccountBackups`

### Additional ideas
- Before starting the actual sync, we could show a `y/N` prompt in the CLI below the list of backups that were found, sort of like a `yum` install.
- Once account/device peers are linked, they should stay synced, buuuut, we may want to run a full sync on some schedule / interval just to help ensure content is available across devices.

sanderpick:
Fixes #534.


sanderpick:
This is cooked... adds the ability to join all your threads from a new peer. Take a look when you guys get a chance!

core/overview.go
```diff
@@ -6,24 +6,28 @@ import (
...
 	AccountPeerCount int `json:"account_peer_cnt"`
 	ThreadCount      int `json:"thread_cnt"`
 	FileCount        int `json:"file_cnt"`
 	ContactCount     int `json:"contact_cnt"`
 }
 
-// Overview returns an overview object
-func (t *Textile) Overview() (*Overview, error) {
+// Summary returns a summary of node data
+func (t *Textile) Summary() (*Summary, error) {
```
carsonfarmer: Is this exposed via the REST API? If not, it defs should be. Would be super useful to have.

core/search.go
```diff
@@ -60,9 +61,18 @@ func (s *queryResultSet) List() []*pb.QueryResult {
 	return list
 }
 
+// Full returns whether or not the number of results meets or exceeds the
```
carsonfarmer: Comment got cutoff here?

sanderpick: good eye!

core/overview.go
sanderpick: oh yeah, should be for sure

PR #544, opened 2019-02-19
Generate and release js types

asutula:
For now, I'm just integrating this into the existing release flow (ie, doing it how we do the frameworks), so the types end up as downloadable from github releases and and downloaded into an npm-installed `go-mobile` module.


asutula:
Ok after some flailing around testing that the js types build works, this is now g2g.

.circleci/config.yml
```diff
@@ -517,6 +558,7 @@ jobs:
           # mv ~/desktop/* ~/dist/
           mv ~/ios_framework/* ~/dist/
           mv ~/android_aar/* ~/dist/
+          mv ~/js_types/* ~/dist/
```
asutula: I noticed, I don't think these `mv` commands are doing anything. We're already outputting all the artifacts in the various build steps (the cli, ios framework, android aar, js types) to `~/dist/`.

asutula: Guess I'm just mentioning it here to get a second opinion.

sanderpick: Can't remember exactly what's going on there, but I think it's just "flattening" the dist folder... so like all the artifacts in cli are moved into the top level of dist

PR #545, opened 2019-02-20
Export js module that can be accessed externally

asutula:
This is necessary to be able to consume the generated JS types easily. Not the ideal setup because I'm adding the index file to this repo while we still download the generated code that the index depends on, but we can clean that up when we automate the release.


sanderpick:
Ah yeah, makes sense... can you bump common/version.go to rc41?

asutula:
Done

PR #546, opened 2019-02-20
Adds full swagger-based docs

carsonfarmer:
No major code changes here, just tons of docs! There are bound to be copy-pasta mistakes in here... so feel free to nit, or pick some of your fav to double check. Then we can just create tickets for issues as they crop up.


Makefile
```diff
@@ -22,6 +22,7 @@ lint:
 	golint `go list ./... | grep -v /vendor/`
 
 build:
+	swag init -g core/api.go
```
carsonfarmer: Now it'll automatically rebuild docs, which doesn't take a ton of extra time, but we _might_ want to think about having this in a separate make target?

core/api_contacts.go
```diff
@@ -23,6 +36,10 @@ func (a *api) addContacts(g *gin.Context) {
 		g.String(http.StatusBadRequest, "invalid contact")
 		return
 	}
+	if contact.Id != g.Param("id") {
```
carsonfarmer: as per separate conversation

core/api_invites.go
```diff
@@ -10,6 +10,19 @@ import (
 	"github.com/mr-tron/base58/base58"
 )
 
+type InviteInfo map[string]string
```
carsonfarmer: annoying quirk of swaggo... doesn't allow you to have 'generic' return params

core/api_logs.go
```diff
@@ -9,6 +9,20 @@ import (
 	"github.com/gin-gonic/gin"
 )
 
+type SubsystemInfo map[string]string
```
carsonfarmer: same quirk again

core/api_threads.go
```diff
@@ -114,6 +135,11 @@ func (a *api) addOrUpdateThreads(g *gin.Context) {
 		return
 	}
 
+	if thrd.Id != g.Param("id") {
```
carsonfarmer: as per separate convo

carsonfarmer:
When reviewing this one @sanderpick, let me know if you think we should make serving the local `docs` endpoint optional? Additionally, consider the docs build step, and let me know if we should move that to its own call and step?

Makefile
sanderpick: seems like a good idea to have it separate, build is for the binary

sanderpick:
Damn, lots of docs in there! I'm a little concerned w/ having to maintain the descriptions between the the REST API, go client, and all future clients... thoughts?

sanderpick:
Personally, I would keep the descriptions in the REST API dead simple, if any at all... and leave the detailed descriptions to the client tooling, which is really where folks should be poking around. I'm not suggesting you delete stuff now, I just didn't realize your approach here.

sanderpick:
For the docs endpoint, possible to make it a flag on daemon start, like `--swagger` or something?

carsonfarmer:
Re: too much description, yes this is a valid concern. Didn't occur to me until I was deep into it, but valid point. We can actually mostly 'drop' the 'description' option and stick with just 'summary'.

carsonfarmer:
Re: `--swagger` flag, yup, easy to add this. Could be nice. What makes more sense, `--docs` or `--swagger` (just wondering if swagger is less informative than docs).

sanderpick:
Ah yeah, `--docs` seems better

carsonfarmer:
Used `--serve-docs` because `daemon` already has a short code `-d` for `--debug` and didn't want to mess with that. Also moved the docs build code to its own `build_docs` target in `Makefile`.

carsonfarmer:
Suck in some extra fixes there for ya @sanderpick. Should be GTG now.

Makefile
carsonfarmer: Agreed, moved

PR #547, opened 2019-02-20
Include src in package files list

asutula:
For some reason, `mobile/src/index.d.ts` isn't getting included in the npm module. Strangely, `mobile/src/index.js` is. I'm not sure this will fix it, but hoping so. Ideas @sanderpick?

### DIFF
--- a/mobile/package.json
+++ b/mobile/package.json
@@ -25,7 +25,8 @@
     "android"
   ],
   "files": [
-    "scripts"
+    "scripts",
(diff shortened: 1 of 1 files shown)


asutula:
Ah yes, pretty sure that's the problem: https://docs.npmjs.com/files/package.json#files

asutula:
The files in `"main"` gets included no matter what.. that explains `index.js` being included.

PR #553, opened 2019-02-22
Generate JS types using pbjs

asutula:


asutula:
Ok this is g2g and needed to move forward on SDK and Photos update.

PR #562, opened 2019-02-27
docker: adds Dockerfile.cafe

sanderpick:
Adds an additional Dockerfile for use as a cafe... uses server profile, opens cafe, etc..

### DIFF
--- a/textile.go
+++ b/textile.go
@@ -258,9 +258,9 @@ func (x *initCmd) Execute(args []string) error {
 		LogToDisk:       !x.Logs.NoFiles,
 		Debug:           x.Logs.Debug,
 		CafeOpen:        x.CafeOptions.Open,
-		CafePublicIP:    x.CafeOptions.PublicIP,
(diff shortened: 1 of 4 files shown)


textile.go
```diff
@@ -258,9 +258,9 @@ func (x *initCmd) Execute(args []string) error {
 		LogToDisk:       !x.Logs.NoFiles,
 		Debug:           x.Logs.Debug,
 		CafeOpen:        x.CafeOptions.Open,
-		CafePublicIP:    x.CafeOptions.PublicIP,
-		CafeURL:         x.CafeOptions.URL,
-		CafeNeighborURL: x.CafeOptions.NeighborURL,
+		CafePublicIP:    envOrFlag("CAFE_HOST_PUBLIC_IP", x.CafeOptions.PublicIP),
```
sanderpick: Allows these options to be passed as ENV vars, which is needed to get through docker run.

PR #570, opened 2019-02-28
Allow built-in schemas and fix null pointer issue

carsonfarmer:
Quick fix to allow built-in schemas to be specified (i.e., `--media`). Also addresses nul pointer issue which results from move to protobufs.

### DIFF
--- a/core/api_threads.go
+++ b/core/api_threads.go
@@ -51,7 +51,9 @@ func (a *api) addThreads(g *gin.Context) {
 	}
 
 	if opts["schema"] != "" {
-		config.Schema.Id = opts["schema"]
+		config.Schema = &pb.AddThreadConfig_Schema{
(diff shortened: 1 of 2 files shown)


core/api_threads.go
```diff
@@ -51,7 +51,9 @@ func (a *api) addThreads(g *gin.Context) {
 	}
 
 	if opts["schema"] != "" {
-		config.Schema.Id = opts["schema"]
+		config.Schema = &pb.AddThreadConfig_Schema{
```
carsonfarmer: Was giving me issues because I think if `Schema` hasn't been initialized yet, trying to assign to `Id` panics. Pls check to make sure this is right (seems to work)?

sanderpick:
Good find @carsonfarmer!

PR #572, opened 2019-02-28
Minimal desktop

carsonfarmer:
This rips out most desktop functionality in favor of exposing the REST API via a simple tray app. Notifications should still show up using the os notifications, and there are tools to access features of the running daemon. Also re-enables desktop releases, and changes the underlying cors lib we're using to something better maintained and powerful, but simpler to use. Comments will be added inline shortly.


core/api.go
```diff
@@ -8,22 +8,25 @@ import (
...
-	"github.com/gin-contrib/cors"
+	"github.com/textileio/textile-go/pb"
+
+	"github.com/golang/protobuf/proto"
+
 	limit "github.com/gin-contrib/size"
 	"github.com/gin-gonic/gin"
 	"github.com/golang/protobuf/jsonpb"
-	"github.com/golang/protobuf/proto"
+	cors "github.com/rs/cors/wrapper/gin"
```
carsonfarmer: swapping in a difference cors lib. this one is better maintained, simpler to use, and supports wildcard usage which makes cors on localhost easier/better

core/api.go
```diff
@@ -8,22 +8,25 @@ import (
...
-	"github.com/textileio/go-textile/common"
-	_ "github.com/textileio/go-textile/docs"
-	m "github.com/textileio/go-textile/mill"
-	"github.com/textileio/go-textile/pb"
-	"github.com/textileio/go-textile/repo/config"
+	"github.com/textileio/textile-go/common"
+	m "github.com/textileio/textile-go/mill"
+	"github.com/textileio/textile-go/repo/config"
+
+	// blank import for server api docs
```
carsonfarmer: linter was complaining about the blank import, this fixes that

core/api.go
```diff
@@ -407,37 +410,23 @@ func (a *api) abort500(g *gin.Context, err error) {
 }
 
 // getCORSSettings returns custom CORS settings given HTTPHeaders config options
-func getCORSSettings(config *config.Config) cors.Config {
+func getCORSSettings(config *config.Config) cors.Options {
```
carsonfarmer: due to change in cors lib

desktop/Makefile
```diff
@@ -0,0 +1,5 @@
+bundle:
```
carsonfarmer: don't really need this, lemme know if you want it dropped

desktop/bundler.json
```diff
@@ -0,0 +1,26 @@
+{
```
carsonfarmer: makes it easier to do all archs at once

repo/config/init.go
```diff
@@ -120,7 +120,10 @@ func Init() (*Config, error) {
 					"X-Textile-Opts",
 					"X-Requested-With",
 				},
-				"Access-Control-Allow-Origin": {},
+				"Access-Control-Allow-Origin": {
```
carsonfarmer: another option here would be [], which then means all origins are ok. honestly not a bit deal given cors is really only a browser thing anyway... but this feels a bit more 'locked down' than all origins

carsonfarmer:
Once that passes, this is GTG @sanderpick!

desktop/main.go
```diff
@@ -182,137 +86,208 @@ func start(a *astilectron.Astilectron, w []*astilectron.Window, _ *astilectron.M
...
+	repoPath := filepath.Join(appDir, "repo")
+
+	// run init if needed
+	if !fsrepo.IsInitialized(repoPath) {
+		accnt := keypair.Random()
+		initc := core.InitConfig{
+			Account:   accnt,
+			RepoPath:  repoPath,
+			LogToDisk: true,
+			GatewayAddr: fmt.Sprintf("127.0.0.1:5052"),
```
carsonfarmer: Had these as random ports at one point, switched back to fixed so its easier to have 'defaults' for clients. Worth a chat though to make sure this is a good idea @sanderpick .

sanderpick: sounds good to me

.circleci/config.yml
```diff
@@ -583,7 +583,7 @@ jobs:
         command: |
           mkdir -p ~/dist
           mv ~/cli/* ~/dist/
-          # mv ~/desktop/* ~/dist/
+          mv ~/desktop/* ~/dist/
```
sanderpick: sweet!

README.md
```diff
@@ -245,6 +245,37 @@ There are various things to build…
...
+You can also build the architecture-specific versions with:
+
+```
+astilectron-bundle -v -c bundler.{darwin,linux,windows}.json
+```
+
+##### Linux
+
+On Linux, you also have to `apt-get install libappindicator1 xclip libgconf-2-4` due to an issue with building Electron-based apps.
+
```
sanderpick: awesome

andrewxhill: Ah... okay, I was expecting the Makefile in desktop/ to still be the spot to look. Should we remove that now?

PR #602, opened 2019-03-06
cafe: adds deregistration handling

### LINKED ISSUE #542: Cafe Unregistration
Right now when a client disconnects from a cafe, they're:
- Announcing to threads their new set of inboxes (w/o the cafe that was removed)
- Deleting the local cafe session

They also need to send an "unregister" message so the cafe can delete:
- Client thread backups
- Client messages
- The client

sanderpick:
Fixes #542.


sanderpick:
Cheers @carsonfarmer

Issue #540, opened 2019-02-18
Use account thread to track Contacts


sanderpick:
So that contacts are synced between account peers.

Issue #485, opened 2019-01-25
Rename a thread


sanderpick:
We'll need to announce the thread name to the members as well as updating the local index.

sanderpick:
Client API additions:

CLI: 
```
textile threads rename <name> -t <thread_id>
```

HTTP REST API:
```
// renameThreads godoc
// @Summary Rename a thread
// @Description Renames a thread. Only initiators can rename a thread.
// @Tags threads
// @Produce application/json
// @Param id path string true "id"
// @Param X-Textile-Args header string true "name"
// @Success 200 {string} string "ok"
// @Failure 400 {string} string "Bad Request"
// @Router /threads/{id}/name [put]
```

Mobile:
```
func (m *Mobile) RenameThread(id string, name string) error
```

PR #607, opened 2019-03-09
account: peer<->peer thread control (join/leave)

### LINKED ISSUE #539: Account peer<->peer thread control messages
Account peers can tell each other what to do:
- Create a thread
- Remove a thread

sanderpick:
Originally, I had gone down a more complicated path. The approach I landed on in the end is pretty simple. Based on a new (additional) signature added to `ThreadEnvelope`, account peers can control each others' thread joins and leaves. This also leaves the door open to more control via other messages in the future.

Fixes #539.


sanderpick:
Snuck #540 in here too.

sanderpick:
Snuck #485 (Rename a thread) in here too.

PR #616, opened 2019-03-13
pb: Add package specifiers

asutula:
Needed so the generated Java classes have a package. Also added the proto package as well because it's a good practice.


andrewxhill:
how should I be importing now?

getting this on SDK when simply updating to the new release version

Screen Shot 2019-03-14 at 7 32 14 PM

PR #628, opened 2019-03-14
Handle multiple peers for the same account address

### LINKED ISSUE #531: Invite command should not require contact to be known
If not known, look for the contact via search.

### LINKED ISSUE #565: Handle multiple contacts for same account
As we move toward multi-device, we'll need to start dealing with contacts from the same account. We'll want to still keep the contact object tied to a single peer, but we can roll up multiple into something consumable that contains all peers and inboxes.

sanderpick:
Contacts -> Peers, Cards -> Contacts

I think this makes a lot more sense now. `Peer`s map to IPFS peers. A `Contact` can contain multiple peers.

Fixes #565 
Fixes #531


sanderpick:
TODO:
- [ ] Build contacts from search results
- [ ] Finish #531

sanderpick:
Feeling good about this one. Sadly, the multi-contact thing required some actual refactoring. I tried two quick hacks (which ended up wasting time) before biting the bullet and addressing it head on. Now there is a clear distinction between `Contact`s and `Peer`s. However, the API only exposes `Contact`s.

sanderpick:
Objections to merge? Hard to properly review this guy cause of all the small changes, but the deets are covered in the targeted issues: #565, #531

PR #629, opened 2019-03-15
mobile: allow prep files with raw bytes

### LINKED ISSUE #610: Prepare files on mobile should take bytes as well as path
Maybe move the path-based version to another method:

`PrepareFiles`
`PrepareFilesByPath`

### LINKED ISSUE #630: Protobuf package specifiers incorrectly prefix enums
This extends the slack conversation around the proto package specifiers. To be clear, what @asutula did with the `package textile;` should work fine along with the `option go_package = "pb";` and not require any golang code changes... but there seems to be a bug where the proto tags for enum types are prefixed with the package when they shouldn't be. For example (see `Filter`):

```
type QueryOptions struct {
	Local                bool                    `protobuf:"varint,1,opt,name=local,proto3" json:"local,omitempty"`
	Limit                int32                   `protobuf:"varint,2,opt,name=limit,proto3" json:"limit,omitempty"`
	Wait                 int32                   `protobuf:"varint,3,opt,name=wait,proto3" json:"wait,omitempty"`
	Filter               QueryOptions_FilterType `protobuf:"varint,4,opt,name=filter,proto3,enum=textile.QueryOptions_FilterType" json:"filter,omitempty"`
	Exclude              []string                `protobuf:"bytes,5,rep,name=exclude,proto3" json:"exclude,omitempty"`
	XXX_NoUnkeyedLiteral struct{}                `json:"-"`
	XXX_unrecognized     []byte                  `json:"-"`
	XXX_sizecache        int32                   `json:"-"`
}
```

Note: This has nothing to do with remote cafes, as a incorrectly stated earlier.

So, when unmarshaling type with enums, you see error like `mismatched type, expected QueryOptions but got textile.QueryOptions`.

We talked about making all the packages `tx` but I think I'm gonna walk that back because it's a big change for not much gain, and thinking more on it, I think people will understand / recognize the `pb` prefix as something related to models just fine. Also, I'd like to keep the go package directory `pb` and not `tx`.

Sound ok?

sanderpick:
Fixes #610 
Fixes #630


sanderpick:
Old:
- `PrepareFiles`
- `PrepareFilesAsync`

New:
- `PrepareFiles`
- `PrepareFilesSync`
- `PrepareFilesByPath`
- `PrepareFilesByPathSync`

PR #633, opened 2019-03-16
mobile: prepareFiles can take base64

sanderpick:


schema/textile/blob.go
```diff
@@ -0,0 +1,14 @@
+package textile
+
+var Blob = `
```
sanderpick: BLOB preset bonus

PR #659, opened 2019-03-30
threads: fix auto-leave from account peer

sanderpick:
This is purely cosmetic. Thread deletion was happening before processing the leave update from account peers in full, so they would rediscover blocks. However, they weren't added again... just pushed over the update channel.


sanderpick:
Also:
- Cleans up API docs for no-content methods
- Adds `--wait` flag to the `qa/wallet.sh` program

Issue #568, opened 2019-02-27
Switch cafe service back to direct p2p


sanderpick:
With the recent `go-libp2p-kad-dht` dialing fixes (https://github.com/libp2p/go-libp2p-kad-dht/pull/237), we can stop using the cafe service HTTP wrapper. Being able to add a cafe by peer ID is much nice than by URL and allows for a simpler cafe setup.

PR #668, opened 2019-04-04
Cafe Request Handler

### LINKED ISSUE #591: Expose CafeOutbox to native mobile SDKs
Throughout the go lib, methods can add "store requests" which are queued and processed by the `CafeOutbox`. A store request asks a cafe to pin something. On mobile, these requests should all be handled by the OS via background tasks. This would remove the distinction between "things that can be pinned in go land, and things that must be pinned by the OS".

Additional, if we allowed a "store request" to be associated with a file index from an add (normally a store request is just arbitrary data), we could simplify the proposed flow here: https://github.com/textileio/go-textile/issues/586, as the "release" could happen automatically when a queue item is done processing. This would _greatly_ reduce the complexity around adding files on mobile because the SDK won't need to keep track of what's pinned / what's not.. it just works it's way through the upload queue, not caring about anything else.

### LINKED ISSUE #661: Don't modify the system with "make install"
**Describe the bug**

The command `make install` modifies `/usr/local/bin/textile`, which is out of the scope of the user's home directory and shouldn't be modified during development.

**To Reproduce**

1. `make install`

**Expected behavior**

The program should be installed at `~/go/bin/textile`

**Screenshots**

```log
mv dist/textile /usr/local/bin
mv: inter-device move failed: 'dist/textile' to '/usr/local/bin/textile'; unable to remove target: Permission denied
Makefile:25: recipe for target 'install' failed
make: *** [install] Error 1
```

**Additional context**

sanderpick:
Fixes #591 
Fixes #661


sanderpick:
Moving this original info to a comment:

The general idea here is to expose the `pb.CafeRequest`s to a mobile API so that the native OS can process them accordingly.

Major ideas:
- The mobile OS should implement the `cafe.RequestHandler` interface, which requires methods for processing different types of requests
- The internal `CafeService` should also implement this interface, and it should do so w/ libp2p. This means #568.
- In order for others to handle cafe requests over HTTP, the Cafe HTTP API should be formalized / expanded beyond the temporary `/service` handler
- Because some methods are handled very differently over libp2p, the interface has to be pretty "light", e.g., `STORE` requests over p2p first kick off a request to see if the data is even needed (it might be already stored), and then each IPFS block pushed over the connection. Over HTTP, we just PUT the raw data behind the block and forget it
- The node should expose a couple new methods for the handler to grab requests
  - list
  - delete
  - statgroup (see below or code)

Some ongoing notes here, just c/p'd from `cafe/main.go`:

```
list: return a batch of requests by date until "next" is empty
complete: mark a single request as complete (does not delete until all in the group are deleted
so that we can know the total data size being handled)
stat a group: returns some info:
  - num total reqs
  - num complete
  - percent complete
  - will return nil if all done since they will have been deleted

Payloads for different requests (expose something like get payload?):
- store: PUT /store/:cid, body => raw object data
- unstore: DELETE /store/:cid, body => none
- store thread: PUT /thread/:id, body => encrypted thread object (snapshot)
- unstore thread: DELETE /thread/:id, body => none
- deliver message: POST /inbox/:pid, body => encrypted message

Other, related methods...
- stat a thread: are there any pending cafe requests (per cafe) / block messages for this thread?
- stat a files block: what percentage of total bytes for the group are complete, per cafe?
- when you add a new cafe, a bunch of new request should be created, which results in a not-as-synced state

Thoughts...

There are two types of things that need to be "stored" and count to users' data usage:
1. thread blocks (requests grouped by thread id)
2. files dags (requests grouped by target id)
(thread snapshots do not count toward data usage, nor do inboxed messages, ?)
```

sanderpick:
This is ready for a look.

Makefile
```diff
@@ -22,7 +22,7 @@ build:
 	mv textile dist/
 
 install:
-	mv dist/textile /usr/local/bin
+	mv dist/textile $$GOPATH/bin
```
carsonfarmer: Saw the convo around this. What is the implication of this? GOPATH is usually added to default search PATH right? So still available via cli easily?

core/cafe_api.go
```diff
@@ -84,9 +73,30 @@ func (c *cafeApi) start() {
 	// v0 routes
 	v0 := router.Group("/cafe/v0")
 	{
-		v0.POST("/pin", c.pin)
+		v0.POST("/pin", c.validateToken, c.pin)
 		v0.POST("/service", c.service)
 	}
+
+	// v1 routes
+	v1 := router.Group("/api/v1")
```
carsonfarmer: Cool!

core/cafe_api.go
```diff
@@ -84,9 +73,30 @@ func (c *cafeApi) start() {
...
+
+	store := v1.Group("/store", c.validateToken)
```
carsonfarmer: Oh nice, that's slick

core/cafe_api_v1.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := c.node.datastore.CafeClientThreads().Delete(id, client.Id); err != nil {
+		c.abort(g, http.StatusInternalServerError, err)
+		return
+	}
+
+	g.Status(http.StatusNoContent)
+}
+
+func (c *cafeApi) deliverMessage(g *gin.Context) {
+	pid := g.Request.Header.Get("X-Textile-Peer")
```
carsonfarmer: This might have to be added to our default headers in config for CORS stuff. Can leave it for now unless it comes up?

ipfs/main.go
```diff
@@ -286,6 +286,23 @@ func GetObjectAtPath(node *core.IpfsNode, pth string) ([]byte, error) {
 	return json.Marshal(out)
 }
 
+// StatObjectAtPath returns info about an object
+func StatObjectAtPath(node *core.IpfsNode, pth string) (*iface.ObjectStat, error) {
```
carsonfarmer: Oooh, could be super handy!

core/cafe_api_v1.go
sanderpick: There's a couple, this one has already been in use... do those apply to the cafe api?

Makefile
sanderpick: Best practice says to add it to your `PATH` via bash profile, etc...

core/cafe_api_v1.go
carsonfarmer: Ah no, they don't apply to the Cafe API so far, so all good.

Makefile
carsonfarmer: Yeh, cool, nice.

sanderpick:
Thanks @carsonfarmer! Yeah, seems like having a JS client for the cafe would be slick. If you could somehow obtain a cafe session in the browser, you'd be off to the races.

mobile/mobile.go
```diff
@@ -76,8 +76,9 @@ type MigrateConfig struct {
 
 // RunConfig is used to define run options for a mobile node
 type RunConfig struct {
-	RepoPath string
-	Debug    bool
+	RepoPath          string
+	Debug             bool
+	CafeOutboxHandler core.CafeOutboxHandler
```
asutula: Ok so `CafeOutboxHandler` just specifies that the object needs to implement `Flush()`... Is the idea that `Flush` will be called whenever new `CafeRequest`s are added to the queue?

sanderpick: Yep, exactly.

sanderpick:
Just pushed this guy: https://github.com/textileio/go-textile/pull/668/commits/6fb1e2927e185a59756dcb987ecfa7306d738a20

Need to include the core package with `gomobile bind` in order to access the `CafeOutboxHandler` interface.

PR #671, opened 2019-04-09
search: fix remote-only logic in handle cafe query

### LINKED ISSUE #670: Cafe-based thread snapshot queries are broken
Looks like the `RemoteOnly` query option is causing trouble.

sanderpick:
Fixes #670


sanderpick:
Running on dev

PR #690, opened 2019-04-16
Removes tray app

carsonfarmer:
In favor of the new one over at https://github.com/textileio/desktop. This also `tidy`s out the go modules that were specific to the tray app.


carsonfarmer:
Let's hold off on merging this until the issues with CI are sorted over in github.com/textileio/desktop though.

sanderpick:
Probably in there and I just cant see it, but does this update the CI config?

carsonfarmer:
@sanderpick ci stuff updated. We could also probably update the Docker image because we don't need yarn etc anymore?

carsonfarmer:
Should be GTG now @sanderpick

PR #691, opened 2019-04-17
docs: add @balupton

balupton:


sanderpick:
@balupton - We settled on the norm where once somebody else gives a PR approval, the owner does the actual merging. For this reason, I typically assign myself to my own PRs.

PR #695, opened 2019-04-19
ipfs: enable setting of custom ipns key and timeout

andrewxhill:
- also exposes the PrivKey on a thread so I could use it for the IPNS input key


gateway/main.go
```diff
@@ -181,7 +181,7 @@ func (g *Gateway) profileHandler(c *gin.Context) {
 		return
 	}
 
-	pth, err := ipfs.ResolveIPNS(g.Node.Ipfs(), rootId)
+	pth, err := ipfs.ResolveIPNS(g.Node.Ipfs(), rootId, time.Second * 30)
```
balupton: Should the timeout also be a configuration option? Considering how slow IPNS can be.

andrewxhill: yeah, prob should be if we start using IPNS for anything that makes the gateway endpoint useful. the finding from this work was that it's still not going to work for anything we've go, so backing off it.

PR #699, opened 2019-04-23
Another mishmash of tweaks

sanderpick:
The commits say it all.


sanderpick:
Can I get a quick look here?

PR #703, opened 2019-04-24
ci: add editorconfig and prettier

balupton:
Non-functional changes have been applied automatically via editorconfig and prettier, enforcing consistent indentation, line endings, and in some cases code cleanups.

Non-functional changes are:

- Added package.json to ensure consistent prettier version is used between us
- Added installation of prettier via `npm install` to `make setup`
- Added prettier to `make {fmt,lint}`

This knocks off these meta issues for this repository:

- https://github.com/textileio/meta/issues/4
- https://github.com/textileio/meta/issues/3

### DIFF
--- a/qa/wallet.sh
+++ b/qa/wallet.sh
@@ -2,47 +2,49 @@
 
 set -e
 
-for i in "$@"
-do
-case ${i} in
-    -n=*|--num=*)
-    num="${i#*=}"
-    shift
-    ;;
-    --wait)
-    wait=yes
-    shift
-    ;;
-    *)
-    ;;
-esac
+for i in "$@"; do
+	case ${i} in
+	-n=* | --num=*)
+		num="${i#*=}"
+		shift
+		;;
+	--wait)
+		wait=yes
+		shift
+		;;
+	*) ;;
+
+	esac
 done
 
-[[ -z "$num" ]] && { echo "Please specify the number of account peers, e.g., -n=3" ; exit 1; }
+[[ -z "$num" ]] && {
+	echo "Please specify the number of account peers, e.g., -n=3"
+	exit 1
+}
 
 declare -a ports
 
 function getPort() {
-    echo $(( ((RANDOM<<15)|RANDOM) % 48128 + 1024 ))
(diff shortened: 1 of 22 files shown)


qa/wallet.sh
```diff
@@ -115,132 +116,130 @@ tpc0=$(echo ${test0} | jq ".peer_count")
...
+	export API=$(getApi i)
+	textile blocks ls -t $(textile config Account.Thread) -l 100 -d | dot -Tpng -o /tmp/peer${i}_account_blocks.png
+	textile blocks ls -t ${thread} -l 100 -d | dot -Tpng -o /tmp/peer${i}_thread_blocks.png
 done
 
 echo "Success."
 
-if [[ "$wait" = "yes" ]]; then
-    wait
+if [[ "$wait" == "yes" ]]; then
```
balupton: Seems like it caught a bug here.

sanderpick: https://stackoverflow.com/a/2237103/641834

sanderpick: New conventions? It was working fine... maybe not in newer shells.

balupton: I think we may need to adopt https://github.com/textileio/meta/issues/2 as the different shell scripts in this repo are using different shells.

This particular file is using bash, despite the `.sh` extension.

Bash treats `=` and `==` the same within square brackets, so the formatter swapped it for the option that reduces user error I guess.

So this is a non-functional change as expected.

We should take steps though to ensure our shell scripts are consistent, ref:

- https://github.com/textileio/meta/issues/5
- https://github.com/textileio/meta/issues/2
- https://github.com/textileio/meta/issues/1

sanderpick: Totally. Wow, I'm sold on the formatter.

balupton:
@sanderpick once merged, you should add this plugin to your code editor:

- https://plugins.jetbrains.com/plugin/10456-prettier

@carsonfarmer @asutula @andrewxhill for us VSCode users, make sure you have this plugin installed:

- https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

PR #721, opened 2019-05-01
Avatar and contact improvements

### LINKED ISSUE #466: Use file path for setting avatar via cli client / mobile
**Is your feature request related to a problem? Please describe.**
Currently, `textile profile set avatar` only accepts a locally indexed image hash, this isn't very easy to do from the cli..

**Describe the solution you'd like**
It would be useful to take an actual file path.

**Describe alternatives you've considered**
Alternatives include taking a non-locally indexed hash, or easier tooling to access local file hashes. This the above proposed solution seem like the most intuitive to a dev/user.

**Additional context**
This will speed up testing, and make it easier to use cli peers for testing things like avatar display etc.

### LINKED ISSUE #672: Missing name/avatar on account sync
**Describe the bug**
When syncing accounts between devices, if an existing account peer already has an avatar and/or username, these do not seem to be maintained after a new account peer is announced. This may be due to how we are handling updated peer information in `GetBest` on the `PeerDB`. It currently naively gets the most recently updated peer for the requested account... which _might_ not have an avatar or name yet.

**Expected behavior**
Ideally, there would be a bit more logic there to add those potentially missing properties from other existing peers if present, while keeping the new info from the latest update.

### LINKED ISSUE #687: Avatar thread should just be the account thread
Currently, the account thread doesn't have a schema... may as well use the avatar schema and use it instead of creating a new one. This way we can have a single avatar schema per account.

sanderpick:
Fixes #687.
Fixes #466.
Fixes #672.

### DIFF
--- a/cmd/profile.go
+++ b/cmd/profile.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"errors"
 
+	"github.com/textileio/go-textile/keypair"
 	"github.com/textileio/go-textile/pb"
 )
 
@@ -108,10 +109,28 @@ Sets the peer avatar from an image path (JPEG, PNG, or GIF).`
 
 func (x *setAvatarCmd) Execute(args []string) error {
 	setApi(x.Client)
-	if len(args) == 0 {
-		return errMissingAvatar
+
+	_, contact, err := callGetAccount()
+	if err != nil {
+		return err
 	}
-	res, err := executeStringCmd(POST, "profile/avatar", params{args: args})
+	kp, err := keypair.Parse(contact.Address)
+	if err != nil {
+		return err
+	}
+	id, err := kp.Id()
+	if err != nil {
+		return err
+	}
+
+	opts := map[string]string{
+		"thread": id.Pretty(),
+	}
+	if err := callAddFiles(args, opts); err != nil {
+		return err
+	}
+
+	res, err := executeStringCmd(POST, "profile/avatar", params{})
 	if err != nil {
 		return err
 	}
--- a/core/contacts.go
+++ b/core/contacts.go
@@ -109,7 +109,7 @@ func (t *Textile) SearchContacts(query *pb.ContactQuery, options *pb.QueryOption
 				}
 				contacts[peer.Address].Peers = append(contacts[peer.Address].Peers, peer)
 
-				value, err := proto.Marshal(contacts[peer.Address])
+				value, err := proto.Marshal(ensureContactUser(contacts[peer.Address]))
 				if err != nil {
 					terrCh <- err
 					break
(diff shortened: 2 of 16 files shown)


cmd/profile.go
```diff
@@ -108,10 +109,28 @@ Sets the peer avatar from an image path (JPEG, PNG, or GIF).`
 
 func (x *setAvatarCmd) Execute(args []string) error {
```
sanderpick: The command-line client now leverages the add files method using the account thread, and then hits the profile/avatar endpoint which just announces the change to all threads.

core/contacts.go
```diff
@@ -190,5 +190,24 @@ func (t *Textile) contactView(contact *pb.Contact, addThreads bool) *pb.Contact
 		}
 	}
 
+	return ensureContactUser(contact)
+}
+
+// ensureContactUser chooses the "best" user info (name, avatar) from peers to display
+func ensureContactUser(contact *pb.Contact) *pb.Contact {
```
sanderpick: This should fix the avatar and name wipe after a sync.

core/thread.go
```diff
@@ -141,6 +141,16 @@ func (t *Thread) Head() (string, error) {
 	return mod.Head, nil
 }
 
+// LatestFiles returns the most recent files block
+func (t *Thread) LatestFiles() *pb.Block {
```
sanderpick: This new method is used to get the latest files block (avatar) from the account thread, but may have other uses in the future.

core/threads.go
```diff
@@ -496,23 +496,41 @@ func (t *Textile) addAccountThread() error {
 		}
 		// catch malformed account threads from 0.1.10
 		if x.Id == aid.Pretty() {
+
+			// catch schema-less account threads from 0.1.11
+			if x.Schema == nil {
```
sanderpick: Sort of a migration that adds the avatar schema to the account thread if it's missing.

repo/db/peers.go
```diff
@@ -111,18 +111,40 @@ func (c *PeerDB) Get(id string) *pb.Peer {
 	return res[0]
 }
 
-func (c *PeerDB) GetBest(id string) *pb.Peer {
+func (c *PeerDB) GetBestUser(id string) *pb.User {
```
sanderpick: This tries to get the best info for a peer by checking against other peers that have the same address. Related to the post-sync name and avatar wipe of your _own_ contact, this should fix that same issue cropping within the feed.

mobile/profile.go
```diff
@@ -45,12 +45,3 @@ func (m *Mobile) Avatar() (string, error) {
 
 	return m.node.Avatar(), nil
 }
-
-// SetAvatar calls core SetAvatar
-func (m *Mobile) SetAvatar(hash string) error {
```
sanderpick: This is removed because to set the avatar on mobile, you just have to add a file to the account thread (the mobile API handles the announcement by detecting the thread is the account thread). In the future, a `setAvatar` method could be added to the iOS/Android/RN SDKs which would just do that behind the scenes.

core/thread.go
andrewxhill: yeah, nice

repo/db/peers.go
```diff
@@ -111,18 +111,41 @@ func (c *PeerDB) Get(id string) *pb.Peer {
...
+			latest = &pb.User{Address: addr, Name: name, Avatar: avatar}
+		} else if latest != nil {
+			if name != "" && latest.Name != "" {
+				latest.Name = name
+			}
+			if avatar != "" && latest.Avatar != "" {
+				latest.Avatar = avatar
+			}
+			if latest.Name != "" && latest.Avatar != "" {
+				rows.Close()
```
sanderpick: Oop, when returning early from a row scan like this, subsequent calls will block the db connection unless explicitly closed.

PR #726, opened 2019-05-02
docs: add inline docs for textile.go

balupton:
/ref https://github.com/textileio/meta/issues/20

PR #732, opened 2019-05-03
core: prefer flag value over env value

balupton:
resolves https://github.com/textileio/go-textile/pull/726#discussion_r280688442


balupton:
We should document this change in a changelog, as users who where using both, will have different behaviour once this merges.

balupton:
@carsonfarmer this needs to be reviewed again, due to the conflict resolution

Issue #744, opened 2019-05-07
Add PUT and PATCH options to default API Access-Control-Allow-Methods


carsonfarmer:
Getting some CORS errors from the api client when performing contact search and add.

balupton:
@carsonfarmer there doesn't seem any downside to enabling this for all, so seems we can future proof this by just enabling all of them:

```
http.MethodConnect,
http.MethodDelete,
http.MethodGet,
http.MethodHead,
http.MethodOptions,
http.MethodPatch,
http.MethodPost,
http.MethodPut,
http.MethodTrace,
```

carsonfarmer:
Yeh good call... "collect em all"!

PR #747, opened 2019-05-08
core: CORS on Gateway, expand CORS on API

### LINKED ISSUE #692: Add CORS support to gateways
We now have outside developers using our decrypting gateways to fetch data on behalf of users. While this isn't ideal, for now it is the best course of action when building pure browser apps on top of Textile data. To support such usage, we need to add CORS support so that web-apps can `fetch` data via the gateways.
We already have nice CORS support for the REST APIs, so there is a working example to base this work off. For the gateways, things can probably be even simpler, because we can just open up to all domains, which I think is the default for the CORS lib we're using.

balupton:
Closes #692, #744


balupton:
Ready for review.

balupton:
Merged with mater. Now ready for review.

balupton:
The [`.gitignore` changes](https://github.com/textileio/go-textile/pull/747/files#diff-a084b794bc0759e7a6b77810e01874f2) as described by @sanderpick 

> I was adding a testdata folder with a .gitkeep file for your test (the pattern we have for tests put repo dirs in testdata/.textile, relative to the the package dir.
>
> when i was doing that i wanted to make sure we we’re gitignoring testdata folders (we aren’t) and took the opportunity to clean up that file (tray had been moved to its own repo).

repo/config/init.go
```diff
@@ -105,20 +115,31 @@ func Init() (*Config, error) {
...
+					http.MethodTrace,
 				},
 				"Access-Control-Allow-Headers": {
+					// rs/cors default headers
+					"Origin",
+					"Accept",
 					"Content-Type",
+					"X-Requested-With",
+					// reason why this is here is unknown
 					"Method",
```
carsonfarmer: Ha, past copy-pasta error?

balupton: nope, it is there, is a not for me having no idea why `Method` is there

balupton:
> Yes this looks good to me @balupton. I think this has already been seen by @sanderpick correct?

Sander hasn't seen it yet. His comments were for #749

PR #749, opened 2019-05-08
WIP: core: decrypt blocks

### LINKED ISSUE #688: Get raw file from block id
After adding a file with the CLI you get the block id response.

Without having to look at that block and parse out the files, is there a way to get the raw file back?

for example, using the block id `QmWWasyab1KB2xVPYHUtYxHNEa6PjD24w1cjSu8tfn4mMm`

`textile files get QmWWasyab1KB2xVPYHUtYxHNEa6PjD24w1cjSu8tfn4mMm` 

will just return the block. is there a trick the following that i'm missing?

`textile files get QmWWasyab1KB2xVPYHUtYxHNEa6PjD24w1cjSu8tfn4mMm/0`

balupton:
Closes #688

# Status

Not yet ready for review. WIP.

#  Changelog

Files:

- Renames `FileData` to `FileContent`
- Introduces `FileIndexContent` helper.

API:

- `/blocks/:id/` redirects to `/blocks/:id/meta ` for b/c
- `/blocks/:id/meta` returns metadata for the block
- `/blocks/:id/files` returns the files for the block
- `/blocks/:id/files/:index/:path/meta` returns the metadata for the path at the block
- `/blocks/:id/files/:index/:path/content` returns the decrypted content for the path at the block
- `/files/:block` redirects to `/blocks/:id/files`

# Todo

- [ ] Tests
- [ ] Documentation for new API methods
- [ ] Accompanying CLI methods

### DIFF
--- a/core/api.go
+++ b/core/api.go
@@ -177,7 +177,21 @@ func (a *api) Start() {
 
 			block := blocks.Group("/:id")
 			{
-				block.GET("", a.getBlocks)
+				block.GET("/meta", a.getBlockMeta)
+				files := block.Group("/files")
+				{
+					files.GET("", a.getBlockFiles)
+					file := files.Group("/:index/:path")
+					{
+						file.GET("/meta", a.getBlockFileMeta)
+						file.GET("/content", a.getBlockFileContent)
+					}
+				}
+
+				block.GET("", func (g *gin.Context) {
+					id := g.Param("id")
+					g.Redirect(http.StatusPermanentRedirect, "/api/v0/blocks/" + id + "/meta")
+				})
 				block.DELETE("", a.rmBlocks)
 
 				block.GET("/comment", a.getBlockComment)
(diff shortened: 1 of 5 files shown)


core/api.go
```diff
@@ -408,7 +424,7 @@ func (a *api) getFileConfig(g *gin.Context, mill m.Mill, use string, plaintext b
 	} else {
 		var file *pb.FileIndex
 		var err error
-		reader, file, err = a.node.FileData(use)
+		reader, file, err = a.node.FileContent(use)
```
balupton: this should be simplified via `:=` as well

sanderpick: Using `:=` would create a new local var, whereas here we mean the `reader` defined at the beginning of the function.

balupton: Good point.

balupton:
Superseded by https://github.com/textileio/go-textile/pull/776

PR #750, opened 2019-05-08
WIP: adds delete account method

andrewxhill:


core/account.go
```diff
@@ -35,6 +35,20 @@ func (t *Textile) Decrypt(input []byte) ([]byte, error) {
 	return t.account.Decrypt(input)
 }
 
+// DeleteAccount removes the local account data by wiping the repo
+func (t *Textile) DeleteAccount(repoPath string) error {
```
andrewxhill: I'm requiring that the repoPath is supplied at delete time to help insure intent here.

core/account.go
```diff
@@ -35,6 +35,20 @@ func (t *Textile) Decrypt(input []byte) ([]byte, error) {
...
+	if t.Started() {
+		// Node shouldn't be running when delete requested
+		return fmt.Errorf("node is started")
+	}
+	if 0 < len(t.CafeSessions().Items) {
```
andrewxhill: Is the existence of cafesessions enough of a check? or could a cafe be registered with no session that needs to be notified?

core/account.go
sanderpick: Based on the convo from yesterday, I thought we were going down the path where the node manages the filesystem? In which case, you'd just want to pass the account address here and we can use the existing repo path on Textile as a prefix: 

```
<repo_path>/
  <address_a>/...
  <address_b>/...
  <address_c>/...
```

Where a, b, and c do _not_ translate to wallet indexes... these could be from diff wallets.

core/account.go
andrewxhill: cool, yep

core/account.go
andrewxhill: sure. does it make sense then to ensure that the one doing the removing who owns the account. 

```js
	if len(t.Account().Address() != address {
		// Ensure that an authorized user is doing the removal
		return fmt.Errorf("not authorized to remove this account")
	}
```

or is it safe to assume  that this method could just be run by anyone who controls the device without much consequence. As in, maybe you sign in with my device and then sign out, but i want to fully remove your account info from my device later... something like that.

sanderpick: Hmm, perhaps no because it would be pretty superficial (you clearly have access to the filesystem and could simply delete my repo at any time outside of the API).

We should probably back up a little. I think @asutula and I had talked previously about this API being on the mobile Wallet object, as in, it does interop with core Textile. In which case, I spoke too soon regarding the repopath... I guess you'd need both the path (as the prefix) and address in order to delete a repo.

Wdyt of this approach? Sorry for the quick thinking earlier.

andrewxhill: `DeleteAccount(repoPath string, address string)` works for me, makes sense

sanderpick: Cool. Typo fix above, should be: "it does _not_ interop with `core.Textile`"

andrewxhill: Okay so theeen the question becomes, what to do about cafe deregistration. Before I had been thinking that that current account would be the one you'd delete, so

```
	if len(t.CafeSessions().Items) > 0 {
		// Accounts must de-register before deleting
		return fmt.Errorf("node is registered with cafes")
	}
```

however, if you are deleting a different account this check would fail incorrectly perhaps... should we just try to guide a client deregister first and skip the check?

sanderpick: Oop, didn't see this message. Right yeah, you won't even have a reference to `t` if you go full independent wallet API. Gonna move to the wallet api issue w/ more thoughts.

PR #756, opened 2019-05-09
Mobile avatar helper

sanderpick:
This is a little nasty but will clean up nice when I redo the mobile files API.

### DIFF
--- a/mobile/files.go
+++ b/mobile/files.go
@@ -24,19 +24,19 @@ import (
 // PrepareFiles processes base64 encoded data for a thread, but does NOT share it
 func (m *Mobile) PrepareFiles(data string, threadId string, cb Callback) {
 	go func() {
-		cb.Call(m.PrepareFilesSync(data, threadId))
+		cb.Call(m.prepareFilesSync(data, threadId))
 	}()
 }
 
 // PrepareFilesByPath processes a file by path for a thread, but does NOT share it
 func (m *Mobile) PrepareFilesByPath(path string, threadId string, cb Callback) {
 	go func() {
-		cb.Call(m.PrepareFilesByPathSync(path, threadId))
+		cb.Call(m.prepareFilesByPathSync(path, threadId))
 	}()
 }
 
 // PrepareFiles processes base64 encoded data for a thread, but does NOT share it
-func (m *Mobile) PrepareFilesSync(data string, threadId string) ([]byte, error) {
+func (m *Mobile) prepareFilesSync(data string, threadId string) ([]byte, error) {
 	if !m.node.Started() {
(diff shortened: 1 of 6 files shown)


mobile/files.go
```diff
@@ -24,19 +24,19 @@ import (
 // PrepareFiles processes base64 encoded data for a thread, but does NOT share it
 func (m *Mobile) PrepareFiles(data string, threadId string, cb Callback) {
 	go func() {
-		cb.Call(m.PrepareFilesSync(data, threadId))
+		cb.Call(m.prepareFilesSync(data, threadId))
```
sanderpick: @asutula I'm making these sync flavored methods internal.

mobile/mobile_test.go
```diff
@@ -748,6 +747,14 @@ func TestMobile_SetUsername(t *testing.T) {
 	}
 }
 
+func TestMobile_SetAvatar(t *testing.T) {
+	err := mobile1.SetAvatar("../mill/testdata/image.jpeg")
```
sanderpick: Just takes a path. Though you can also add any IPFS hash, or an existing file hash that's needs decryption.

mobile/profile.go
```diff
@@ -47,3 +58,117 @@ func (m *Mobile) Avatar() (string, error) {
...
+	}
+
+	if _, err := thrd.AddFiles(node, "", keys.Files); err != nil {
+		return err
+	}
+
+	return m.node.SetAvatar()
+}
+
+func tmpGetMedia(reader io.Reader) (string, error) {
```
sanderpick: This will get removed when the mobile Files API is simplified.

mobile/profile.go
```diff
@@ -47,3 +58,117 @@ func (m *Mobile) Avatar() (string, error) {
 
 	return m.node.Avatar(), nil
 }
+
+// SetAvatar calls core SetAvatar
+func (m *Mobile) SetAvatar(path string) error {
```
asutula: Looks like this returns nothing or an error. How does remote pinning the data happen? I guess I still expected to have to handle the pin request. Or are you only sending that pin request through the cafe request queue now?

sanderpick: Ahhhh wow. I completely forgot about the mobile pin step. That's why we have the two methods, prepare and add. Once the queue stuff is in place, this will go away. Thinking about how to proceed in the meantime...

sanderpick: We might be best off to just leave the avatar handling as is for now... so you just have to add to the avatar thread via prepare and add.

sanderpick:
Superseded by #762

PR #758, opened 2019-05-10
Improve cafe-less search

sanderpick:
Respond to search query with pubsub if responder is a cafe and the request is not from a cafe.


sanderpick:
Please have a look 👀

PR #770, opened 2019-05-15
first pass on adding contact by add

andrewxhill:
a little post-lunch inspiration, but could use some feedback on right way to handle the search result processing => successful add

### DIFF
--- a/core/contacts.go
+++ b/core/contacts.go
@@ -20,6 +20,29 @@ func (t *Textile) AddContact(card *pb.Contact) error {
 	return nil
 }
 
+// AddContactByAddress adds or updates a card by address
+func (t *Textile) AddContactByAddress(address string) error {
+	resultCh, errCh, _, _ := t.SearchContactsByAddress(address, 10)
+
+	for {
+		select {
+		case err := <-errCh:
+			return err
+
+		case res, ok := <-resultCh:
+			if !ok {
+				return fmt.Errorf("contact not found")
+			}
+			contact := new(pb.Contact)
+			if err := ptypes.UnmarshalAny(res.Value, contact); err != nil {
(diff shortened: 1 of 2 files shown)


GitCop:

 * Commit: 69a1949afd44ff411967f4e0bda517e74bb27973



---

GitCop:

 * Commit: 69a1949afd44ff411967f4e0bda517e74bb27973



---

sanderpick:
Nice dude. This would belong in the mobile client, leaving the one core search contacts method that used the options config.

sanderpick:
I would have tried to explain the challenges here before work commenced... didn’t realize this was part of the sprint. 

The problem is that the searches aren’t cheap and you might want to select multiple contacts from a search to add. So you’d be queueing up a bunch of searches that each have to wait a little bit of time to e sure the latest results arrive. The command line client does this correctly with snapshot (As far as I remember)... it keeps an in mem list that is then available to use directly by the API. I think the right way to go here with contacts would be to do something similar in the mobile API since it only has on user (the app). The rest api can have many users, which is why its clients need to handle this themselves.

andrewxhill:
we can kill it. jumped on it for some personal edification related to search mechanics and go-textile. 

couple things i was thinking putting it together though.

1. put it in core because the same method would be useful exposed in the api. (re your comment, not sure i follow the logic of pushing the legwork out to the client, since the same queuing for multiple add by address (searches) would be required).
2. i scoped it to only be Add by Address (no Add by Name) to try and side step some of that in-mem enhancement need. But maybe I missed the full concept. My thinking was that if you get the result from an Address, you are done, you got it. And in this case it is actually bundled with the add step, so they actually become part of your contact table.

PR #790, opened 2019-05-24
Update to go-ipfs@v0.4.21-rc3

### LINKED ISSUE #786: Random context canceled in tests
Possibly related to the newer go-ipfs 0.4.20+

```
--- FAIL: TestMobile_RemoveThread (0.52s)
    mobile_test.go:394: context canceled
```

Often happens on that test, but I've seen it on others.

### LINKED ISSUE #787: Merges need address reset
From a community member in public slack:

https://github.com/textileio/go-textile/blob/master/core/thread_merges.go#L29

Shouldn't MERGE block be identical not only across one account's peers but also across different users? In sum, should we also reset *Address* there?

### LINKED ISSUE #788: Init flags show for every subcommand
For example, 

```
$ textile threads add --help
usage: textile thread add [<flags>] <name>

Adds and joins a new thread

Flags:
      --help                     Show context-sensitive help (also try --help-long and --help-man).
      --api="http://127.0.0.1:40600"
                                 API Address to use
      --api-version="v0"         API version to use
      --server                   Apply IPFS server profile
      --swarm-ports=SWARM-PORTS  Set the swarm ports (TCP,WS). A random TCP port is chosen by default
      --log-files                If true, writes logs to rolling files, if false, writes logs to stdout
      --debug                    Set the logging level to debug
      --api-bind-addr="127.0.0.1:40600"
                                 Set the local API address
      --cafe-bind-addr="0.0.0.0:40601"
                                 Set the cafe REST API address
      --gateway-bind-addr="127.0.0.1:5050"
                                 Set the IPFS gateway address
      --profile-bind-addr="127.0.0.1:6060"
                                 Set the profiling address
      --cafe-open                Open the p2p Cafe Service for other peers
      --cafe-url=CAFE-URL        Specify a custom URL of this cafe, e.g., https://mycafe.com
      --cafe-neighbor-url=CAFE-NEIGHBOR-URL
                                 Specify the URL of a secondary cafe. Must return cafe info, e.g., via a Gateway:
                                 https://my-gateway.yolo.com/cafe, or a Cafe API: https://my-cafe.yolo.com
      --version                  Show application version.
  -k, --key=KEY                  A locally unique key used by an app to identify this thread on recovery
  -t, --type="private"           Set the thread type to one of: private, read_only, public, open
  -s, --sharing="not_shared"     Set the thread sharing style to one of: not_shared, invite_only, shared
  -w, --whitelist=WHITELIST ...  A contact address. When supplied, the thread will not allow additional peers, useful
                                 for 1-1 chat/file sharing. Can be used multiple times to include multiple contacts
      --schema=SCHEMA            Thread schema ID. Supersedes schema filename
      --schema-file=SCHEMA-FILE  Thread schema filename, supersedes the built-in schema flags
      --blob                     Use the built-in blob schema for generic data
      --camera-roll              Use the built-in camera roll schema
      --media                    Use the built-in media schema

Args:
  <name>  The name to use for the new thread
```

The following flags are only relevant to the init command:

```
--server                   Apply IPFS server profile
      --swarm-ports=SWARM-PORTS  Set the swarm ports (TCP,WS). A random TCP port is chosen by default
      --log-files                If true, writes logs to rolling files, if false, writes logs to stdout
      --debug                    Set the logging level to debug
      --api-bind-addr="127.0.0.1:40600"
                                 Set the local API address
      --cafe-bind-addr="0.0.0.0:40601"
                                 Set the cafe REST API address
      --gateway-bind-addr="127.0.0.1:5050"
                                 Set the IPFS gateway address
      --profile-bind-addr="127.0.0.1:6060"
                                 Set the profiling address
      --cafe-open                Open the p2p Cafe Service for other peers
      --cafe-url=CAFE-URL        Specify a custom URL of this cafe, e.g., https://mycafe.com
      --cafe-neighbor-url=CAFE-NEIGHBOR-URL
                                 Specify the URL of a secondary cafe. Must return cafe info, e.g., via a Gateway:
                                 https://my-gateway.yolo.com/cafe, or a Cafe API: https://my-cafe.yolo.com
```

### LINKED ISSUE #789: `cafes add` command does not take URL
```
$ textile cafes add --token dafvdfv
missing cafe host
```
This should require the cafe URL.

sanderpick:
- Also exposes port 6060 on the docker images for the profiling URL.

Fixes #789
Fixes #788 
Fixes #786 
Fixes #787


cmd/daemon.go
```diff
@@ -2,19 +2,20 @@ package cmd
...
-	"github.com/textileio/go-textile/util"
 	"net/http"
 	"os"
 	"os/signal"
 	"runtime"
 	"runtime/debug"
 	"strconv"
 	"time"
+
+	"github.com/textileio/go-textile/common"
```
sanderpick: @balupton can you setup a `go imports` watcher in go land? This way me just opening a file doesn't cause changes.

sanderpick:
This is running on the dev cafe (`KAMY`). Though anecdotal the mem seems to be lower and stable so far.

cmd/daemon.go
balupton: Should already be configured...

CHANGELOG.md
```diff
@@ -0,0 +1,90 @@
+# go-textile changelog
```
balupton: So I think that we can actually use https://github.com/textileio/go-textile/releases as the changelog. And just link there. There is no reason why it needs to be a `CHANGELOG.md` or `HISTORY.md` file specifically (a convention older than github releases).

cmd/blocks.go
```diff
@@ -89,7 +89,7 @@ func BlockIgnore(blockID string) error {
 }
 
 func BlockFile(blockID string, index int, path string, content bool) error {
-	urlPath := "blocks/"+blockID+"/files"
+	urlPath := "blocks/" + blockID + "/files"
```
balupton: Pity that the go formatters don't seem to enforce this.

cmd/main.go
```diff
@@ -1040,7 +1014,7 @@ func executeBlobCmd(meth method, pth string, pars params) error {
 
 func request(meth method, pth string, pars params) (*http.Response, func(), error) {
 	apiURL := fmt.Sprintf("%s/api/%s/%s", *apiAddr, *apiVersion, pth)
-	if *logDebug {
+	if *daemonLogDebug {
```
balupton: Ok, I should have documented this change.

Before `--debug` existed for daemon command, and init command. I decided to make it a global flag instead, which is common for CLIs, hence this functionality which operates on the CLI client - so doing `textile files get <hash>` will output the request URL that it made.

This functionality will need to be removed, or `--debug` will need to be re-added as a global flag instead.

cmd/daemon.go
sanderpick: Sure thing

core/api.go
```diff
@@ -191,9 +191,9 @@ func (a *api) Start() {
 					}
 				}
 
-				block.GET("", func (g *gin.Context) {
+				block.GET("", func(g *gin.Context) {
 					id := g.Param("id")
-					g.Redirect(http.StatusPermanentRedirect, "/api/v0/blocks/" + id + "/meta")
+					g.Redirect(http.StatusPermanentRedirect, "/api/v0/blocks/"+id+"/meta")
```
balupton: Definitely a pity that the go formatters don't seem to enforce this.

cmd/blocks.go
sanderpick: They do, they just hadn't been run. All these little changes are the result of `make fmt`

cmd/main.go
sanderpick: Ah, nice. I like that. Will change back.

CHANGELOG.md
sanderpick: Cool, I'm just copying go-ipfs here, but agreed, it does feels redundant... _unless_ we can hook into this for the release automation - pull the notes from a `CHANGELOG.md` file and dump into the release notes.

sanderpick:
Ok, changes pushed.

sanderpick:
Added notes to release: https://github.com/textileio/go-textile/releases/tag/v0.2.2

This is deploying to the prod cafes.

PR #792, opened 2019-05-27
Textile.checksum rename arg willEncrypt -> wontEncrypt

### LINKED ISSUE #791: t.checksum arg typo
I'm talking about this 2 lines
https://github.com/textileio/go-textile/blob/master/core/files.go#L48
https://github.com/textileio/go-textile/blob/master/core/files.go#L67

According to the arg name(`willEncrypt`), it should be `!conf.Plaintext` instead.

BTW, as I got the purpose of this `add` suffix inside the `checksum` func is to be able to store both encrypted and unencrypted versions of the same file(and don't reuse the file id between them)

So for backward compatibility, it will be easier to just rename the arg to `wontEncrypt`

requilence:
Need to change the arg name because it is already used with negation bool
fixes #791

### DIFF
--- a/core/files.go
+++ b/core/files.go
@@ -335,9 +335,9 @@ func (t *Textile) fileIndexForPair(pair ipld.Node) (*pb.FileIndex, error) {
 	return t.datastore.Files().Get(c.Cid.Hash().B58String()), nil
 }
 
-func (t *Textile) checksum(plaintext []byte, willEncrypt bool) string {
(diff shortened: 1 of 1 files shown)


GitCop:

 * Commit: f293f672f98a449c755dca6c1a74be3a04f6b948



---

sanderpick:
@requilence would you mind force pushing with a signed commit?

requilence:
> @requilence would you mind force pushing with a signed commit?

Dobe

PR #806, opened 2019-06-06
Update go-ipfs to v0.4.21 + discovery patch

sanderpick:
Re: https://github.com/ipfs/go-ipfs/pull/6415

Also,
- Updates sender handling with the version of `IpfsDHT` used in v0.4.21 (v0.0.13)
- Turns auto relay back on for mobile


repo/config/init_ipfs.go
```diff
@@ -146,7 +148,7 @@ func InitIpfs(identity native.Identity, mobile bool, server bool) (*native.Confi
 			},
 			DisableBandwidthMetrics: mobile,
 			DisableNatPortMap:       server,
-			EnableRelayHop:          false,
+			EnableRelayHop:          server,
```
carsonfarmer: Just double checking, we only want this enabled when run in 'server' mode?

sanderpick: I think so... maybe on desktop but probably requires not being behind a NAT

PR #810, opened 2019-06-06
meta: docs should be generated by CI, not by us

balupton:
Ref https://github.com/textileio/meta/issues/6

Changes:

- delete `docs` directory
- update `.gitignore` with that from textileio/base which ignored the `docs` directory

Signed-off-by: Benjamin Lupton <b@lupton.cc>


balupton:
Review of this will need to let me know what things `docs` affect.

There is also another part of this PR, which will be to amend the git history to truncate the `docs` folder so our repo size is shrunk down. However, that will break all unpushed changes. So not sure when we can schedule that.

carsonfarmer:
`docs.go` needs to be available at build time, as do the various swagger files. These are also needed at run time to serve the local docs. So they probably _do_ need to be in VC? Otherwise, we'd need to re-build them each time? (these aren't just external docs, these are served via `textile daemon --serve-docs`

balupton:
Ok cool.

Rebuilding each time, or at least during publishing, makes sense, as otherwise they risk getting out of date - which is already the current situation

carsonfarmer:
They are an actual go 'package' imported into `core/api.go`. So I _guess_ they'd have to be built each time? That doesn't seem ideal?

balupton:
> Could be added as a step to `make build`? But I think go might have issues with that when `go get`ing `go-textile`, we'll have to test.

good point, more research is needed

PR #762, opened 2019-05-12
New Files API and Cafe Request Handler

sanderpick:
Fixes #737  
Fixes #765 
Fixes #766 
Fixes #760 
Fixes #761 
Fixes #769 
Fixes #573
Fixes #793
Fixes #568 

### Files
- Allow adding data to a thread from an arbitrary IPFS hash. If this hash corresponds to indexed encrypted data, that will be used. Otherwise, the mill input is catted from the network.

### Search
- Adds a standalone search endpoint to the cafe v1 HTTP API (this should eventually move to p2p)

### Cafes
- Moves most cafe communication to p2p
- More intelligent block sync grouping
- Splits pinning and posting
- Allows blocks to be created and left in a "pending" state during their and associated files sync before being "released" to the thread and peers
- Adds "historical" sync after a cafe registration
- Unpins files when the associated block is deleted / ignored
- Simplifies and reduces the number of calls to flush
- Adds mobile sync events
- Includes a mobile cafe sync test

### Events
- `WalletUpdate` -> `AccountUpdate`

### Threads v1.1
In order to handle above sync logic, block parents had to be pulled out of the block. I took this as an opportunity to move to a simple IPLD structure that paves the way for threads v2. A thread "node" has a link for parents (an index map like `"0" -> CID`, etc.) and a link for the encrypted block. The other advantage here is that blocks that are unreachable do not hold up the traversal. This is a trade-off with keeping the links obfuscated. 

- Blocks are wrapped w/ an IPLD node
- Inbox requests contain the entire node (otherwise the block release would have to somehow wait on another pin). The receiving cafe just pins this, making it more readily available for the recipient.
- More intelligent merging: we simply maintain multiple heads per thread that are joined when an outbound update is created.

### Mobile Cafes API
- Re-attempting p2p-based registration

### Mobile Files API
No more prepare, add hell.

- `AddData`
- `AddFiles` (can add multiple)
- `ShareFiles` (can add multiple)
- `SetAvatar`
- allow raw IPFS hashes as input to `AddFiles` (REST API needs updating, can do separately)

### Tests
- Adds comprehensive sync test to both core (desktop handler) and mobile

PR #811, opened 2019-06-07
inboxing: removed per-peer encryption for inboxed msgs

sanderpick:
Also updates the queue processing to the cleaner approach from https://github.com/textileio/go-textile/pull/762.


core/cafe_service.go
```diff
@@ -1155,6 +1155,14 @@ func (h *CafeService) handleDeliverMessage(pid peer.ID, env *pb.Envelope) (*pb.E
 		return nil, nil
 	}
 
+	if msg.Env != nil {
+		id, err := ipfs.AddData(h.service.Node(), bytes.NewReader(msg.Env), true)
```
sanderpick: Let the cafe actually pin the message here before adding to the client's inbox.

Issue #783, opened 2019-05-23
Consider removal of the concept of default threads


balupton:
Requested by @sanderpick, apparently the default threads haven't really seen use.

PR #812, opened 2019-06-07
WIP cli+api: consistent and intuitive CLI usage

balupton:
Fixes https://github.com/textileio/docs/issues/89

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,8 @@
-# 2018 September 26
-# https://github.com/bevry/base
+# 2019 May 25
+# https://github.com/textileio/base
 
+# Standards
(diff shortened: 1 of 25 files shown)


carsonfarmer:
When this is out of WIP, it would be awesome for downstream clients if the changelog included all the new REST API endpoints 🙏. Especially because all the nice aliases are _slightly_ harder to groc in the source :)

balupton:
> When this is out of WIP, it would be awesome for downstream clients if the changelog included all the new REST API endpoints 🙏

No changes to the REST API in this besides [removal of default thread](https://github.com/textileio/go-textile/issues/783), which shouldn't require any client updates, as no one is probably using it.

PR #815, opened 2019-06-07
Escape single quotes in repo pincode

carsonfarmer:
Signed-off-by: Carson Farmer <carson.farmer@gmail.com>

Question: Do we think we need to include this fix in the migration files? Since we've never really enabled passwords before, we're probably ok? But happy to blast this in there if we want to be on the safe side.


sanderpick:
I guess it just wouldn't have worked in the past with single quotes, so probably safe to bet nobody has a password with them, right?

carsonfarmer:
I think so :)

PR #816, opened 2019-06-08
contacts: fixes best peer selection

### LINKED ISSUE #814: Contact details not syncronized into thread update metadata
1. Launch an app, Photos
2. Ensure you have an Avatar set
3. Create some posts in a Group
4. Change your avatar
5. You'll see that your avatar in the Group threads may still show up with your old contact information.

You can test it further by changing your Display name from a paired daemon. You'll see that the display name will propagate to Drawer, Groups screens. But on the Photo feed (where authors are added into each block data) it will still show your old display name.

Hard refresh, you'll keep seeing the same.

The data we use to diplay here comes from the `IUser` model.

The avatars show in Groups comes from a direct query of the Contacts data (it filters Contacts by those containing the threadId)

sanderpick:
Fixes #814

### DIFF
--- a/core/contacts.go
+++ b/core/contacts.go
@@ -208,10 +208,10 @@ func ensureContactUser(contact *pb.Contact) *pb.Contact {
 		return contact
 	}
 	for _, p := range contact.Peers {
-		if p.Name != "" && contact.Name != "" {
+		if p.Name != "" && contact.Name == "" {
 			contact.Name = p.Name
 		}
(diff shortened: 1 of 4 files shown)


sanderpick:
There you go @andrewxhill - pretty sure it was just a this logical flaw in the best peer selection. Tested on my desktop. Keep in mind, you'll get the migration to repover 14 if you try this locally. Working fine for me so far though.

carsonfarmer:
Lgtm!

PR #819, opened 2019-06-10
Mobile sync tweaks

sanderpick:
Here are the tweaks we talked about @asutula:
- Adds error and error ID to `CafeSyncGroupStatus`
- Makes `WriteCafeRequest` async with `Callback`


sanderpick:
release coming...

PR #821, opened 2019-06-11
cli+api: consistent and intuitive CLI usage

balupton:
- Closes textileio/docs#89
- Closes #783
- Closes #793
- Closes #794
- Closes #795
- Closes #797
- Closes #798
- Closes #799
- Closes #809
- Closes #812

## Internal changes:

- next page functionality moved to helper, and wording changed to be more clear

## User facing changes:

- Removed the concept of default threads from textile
  - Removes thread init config that was only used for configuring the textile default thread ID, which this branch has already removed
- Setup better aliases for files, and set list commands as defaults for their parent, e.g.
  - `textile threads` -> `textile threads list`
  - `textile files` -> `textile files list` -> `textile files list thread`
  - `textile thread files` -> `textile files thread` -> `textile files list thread`
  - `textile block files` -> `textile files block` -> `textile files list block` -> `textile files list block list`
- Fixes fetching file content when only one file was in a block:
  - `textile files block <blockid> --index=0 --path=. --content` is now `textile files block get <blockid> --index=0 --path=. --content`
- Commands that add files now output more meaningful information
- `textile subscribe` is now `textile observe`, same behaviour, clearer name, b/c compat provided
  - and for the HTTP API `/subscribe` is now `/observe`, b/c compat provided
- `textile thread unsubscribe` is now `textile thread abandon`, same behaviour, clearer name, b/c compat provided
- `textile token create` is now `textile token add`, with b/c
- Some flags have been converted to arguments for consistency, new signatures are
  - `textile block list <thread>`
  - `textile chat <thread>`
  - `textile feed [<thread>]`
  - `textile file add <thread> [<path>]`
  - `textile init <account-seed>`
  - `textile invite create <thread>`
  - `textile message add <thread> <body>`
  - `textile message list [<thread>]`
  - `textile observe [<thread>]`
  - `textile thread peer <thread>`
  - `textile thread rename <thread> <name>`
  - `textile thread abandon <thread>`
  - `textile wallet create [<passphrase>]`
  - `textile wallet accounts <mnemonic> [<passphrase>]`
- Some flags have been renamed for terseness:
  - `--repo-path` changed to `--repo`
  - `--pin-code` changed to `--pin`
- `textile --help` is now more compact
- Next page output is now clearer on how to exit
- Improved the documentation for `textile init`, `textile account seed`, `textile wallet`, `textile wallet accounts`, `textile observe`, `textile thread abandon`
- Fix `textile contact rm` not working due to typo (regression from v0.2)
[...]

PR #826, opened 2019-06-13
Adds block downloads and more threads v2

### LINKED ISSUE #264: Index failed blocks w/ retries
Part of the more resilient message handing discussed in #240, will also let a thread retry back-propagation starting at the failed block.

sanderpick:
Closes #264 

### Threads
- Block indexes now have three states:
  - `READY`: payload is downloaded and synced (if created locally)
  - `QUEUED`: waiting to sync
  - `PENDING`: waiting to download
- During the thread traversal, the inner payloads (the encrypted part) are downloaded with a queue (`BlockDownloads`), which are handed in concurrent batches like the other queues, while the thread cursor continues to traverse
- Traversal handles parents concurrently 
- The actual download _could_ be moved to OS-specific background fetch APIs on mobile, similar to sync
- Block nodes now contain two extra links:
  - `target`, pulled out of the encrypted block, enabling arbitrary nesting of data and message blocks
  - `data`, the new home for the data link (previously target)
- With the addition of the extra links, the thread graph is traversable by IPFS without touching the encrypted payloads 🎉, which paves the way for an OrbitDB-like public thread type, and `js-threads`
- **Fixes an invites bug which caused direct invites to never be sent**

### Thread Service
- Logical cleanup, which is good prep for moving threads to their own repo (`go-threads`)

### Cafe Service
- Fixes a bug in `handleDeliverMessage` where the envelope payload was not being correctly unmarshalled

### Files
- Can now take a target (nested files), but disabled until 0.5.0 release

### Messages
- Can now take a target (nested messages), but disabled until 0.5.0 release

### Comments
- deprecated (replaces by messages w/ a target), but still active until 0.5.0 release

TODO:
- [x] Fix up the CLI and API after #821 (files and messages)
- [x] Fix up the the mobile API after #821 (files and messages)
- [x] Fix `FeedItem`, needs an extra field for `data` links (used to be target)
- [x] Add a datastore migration
- [x] Download thread tails in the background


sanderpick:
Ready for some 👀 here

core/api_files.go
```diff
@@ -78,7 +78,8 @@ func (a *api) addThreadFiles(g *gin.Context) {
 		return
 	}
 
-	hash, err := thrd.AddFiles(node, opts["caption"], keys.Files)
+	// @todo Allow the setting of the target in 0.5.0
+	hash, err := thrd.AddFiles(node, "", opts["caption"], keys.Files)
```
carsonfarmer: Nice! Looking forward to that :)

core/cafes_test.go
```diff
@@ -136,12 +136,12 @@ func TestCore_HandleCafeRequests(t *testing.T) {
 
 	// check if blocks are pinned
 	var blocks []string
-	var targets []string
+	var datas []string
```
carsonfarmer: 🤓

core/invites.go
```diff
@@ -211,40 +211,46 @@ func (t *Textile) handleThreadAdd(plaintext []byte, parents []string) (mh.Multih
...
-		return nil, err
-	}
-	hash, err := thrd.join(author)
+	hash, err := thread.join(block.Header.Author)
 	if err != nil {
 		return nil, err
 	}
+
+	// handle the thread tail in the background
+	go func() {
```
carsonfarmer: nice!

core/main_test.go
```diff
@@ -33,7 +33,7 @@ func TestInitRepo(t *testing.T) {
 	if err := InitRepo(InitConfig{
 		Account:  accnt,
 		RepoPath: vars.repoPath,
-		ApiAddr:  fmt.Sprintf("127.0.0.1:%s", GetRandomPort()),
+		ApiAddr:  fmt.Sprintf("127.0.0.1:9999"),
```
carsonfarmer: We going to do this throughout? Why the change?

core/thread.go
```diff
@@ -184,129 +190,204 @@ func (t *Thread) UpdateSchema(hash string) error {
...
-		}
-		ends, err := t.followParent(hash)
-		if err != nil {
-			log.Warningf("failed to follow parent %s: %s", parent, err)
-			continue
-		}
-		for _, p := range ends {
-			final[p] = struct{}{}
-		}
+		wg.Add(1)
```
carsonfarmer: This is sick!

core/thread.go
```diff
@@ -478,11 +533,29 @@ func (t *Thread) commitNode(block string, additionalParents []string, updateHead
...
 		return nil, err
 	}
-	err = ipfs.AddLinkToDirectory(t.node(), dir, parentsLinkName, pnode.Cid().Hash().B58String())
+	pnodeId := pnode.Cid().Hash().B58String()
+	err = ipfs.AddLinkToDirectory(t.node(), dir, parentsLinkName, pnodeId)
 	if err != nil {
 		return nil, err
 	}
 
+	// add target
```
carsonfarmer: Nice, very straightforward code here too, I like it.

core/invites.go
sanderpick: should really speed up joins - we can do this now because it's no longer necessary to wait for the last discovered leaves in order to properly update the HEAD

core/main_test.go
sanderpick: For some reason, the random ports were causing the api health test to fail sporadically

cmd/files.go
```diff
@@ -536,8 +536,8 @@ func FileIgnore(blockID string) error {
 // ------------------------------------
 // > files key
 
-func FileKeys(targetID string) error {
-	res, err := executeJsonCmd(http.MethodGet, "keys/"+targetID, params{}, nil)
+func FileKeys(dataID string) error {
+	res, err := executeJsonCmd(http.MethodGet, "keys/"+dataID, params{}, nil)
```
balupton: This should be documented in the changelog if this is a value change and not just a variable name change

cmd/main.go
```diff
@@ -387,10 +387,10 @@ Stacks may include:
 	blockFilesCommand(cmds, fileListCmd, []string{"block"})
 
 	// file keys
-	fileKeysCmd := fileCmd.Command("keys", "Shows the encryption keys for each content/meta pair for the given block DAG target").Alias("key")
-	fileKeysTargetID := fileKeysCmd.Arg("block-target", "Block Target ID").Required().String()
+	fileKeysCmd := fileCmd.Command("keys", "Shows the encryption keys for each content/meta pair for the given block DAG").Alias("key")
+	fileKeysDataID := fileKeysCmd.Arg("block-data", "Block Data ID").Required().String()
```
balupton: This should be documented in the changelog if this is a value change and not just a variable name change

PR #831, opened 2019-06-16
[WIP] IPFS Cluster Integration

### LINKED ISSUE #734: Cafe peers should optionally use IPFS Cluster
Right now, an account peer can sync with a cafe peer. In terms of pinning/storage, this means:

- All thread updates are synced (encrypted)
- All thread _content_ (file DAGs) are synced (encrypted if schema dictates)
- Account-related thread (contacts, profile info)

You _can_ register with multiple cafes. However, this is a pretty poor method of achieving replication, putting extra bandwidth and load on (often low-power, mobile clients).

I have spoken briefly with @hsanjuan about embedding cluster in a cafe… Ideally we could support:

- `textile` w/ an IPFS node embedded for very easy config and deployment (this is the current setup)
- `textile` w/ an IPFS cluster embedded, requiring users to at least configure some additional IPFS nodes attached to the cluster
- `textile` w/ only libp2p services that communicates with an IPFS cluster's HTTP API, requiring users to also configure some additional IPFS nodes attached to the cluster

sanderpick:
Adds the ability to sidecar IPFS Cluster on Textile.

## Changelog

_coming_

## Closes
- Closes #734 

## Fixes
- The `--repo` flag was no longer parsing `~` (missed in the kingpin refactor)

TODO
- [ ] Implement `IPFSConnector` (`ipfs/cluster.go`)


balupton:
> * The `--repo` flag was no longer parsing `~` (missed in the kingpin refactor)

In the code diff, where is the code for this change? If we are doing `--repo="~/repo"` that is incorrect behaviour to expand ~ there, the correct way to do that in most shell scripts is  `--repo=~/repo`, or is the path has special characters inside of it, then these work  `--repo=~"/repo"` as well as `--repo="$HOME/repo"`

cmd/main.go
```diff
@@ -1192,5 +1211,18 @@ func getRepo(repo string) (string, error) {
 		}
 		repo = filepath.Join(appDir, "repo")
 	}
-	return repo, nil
+	expanded, err := homedir.Expand(repo)
```
sanderpick: Here you go @balupton ^

sanderpick:
Right, but I'm okay being "incorrect" here in support a better UX. I'll stand corrected the day somebody complains because they actually had a directory called "~" :)

sanderpick:
Ok, @hsanjuan - I've got all the pieces in place here. However, running into an authorization error when calling `SyncAll` (`https://github.com/textileio/go-textile/blob/sander/cluster/cluster/main_test.go#L188`):

```
17:31:31.842 ERROR  p2p-gorpc: error handling RPC: client does not have permissions to this method, service name: Cluster, method name: SyncAllLocal server.go:176
17:31:31.859 DEBUG    cluster: rpc auth error client does not have permissions to this method, service name: Cluster, method name: SyncAllLocal cluster.go:1597
```

If I understand correctly, the RPC auth is based solely on the cluster secret. The are matching in the test-generated config files.

I'm putting this down for now but let me know if you have any quick ideas...

sanderpick:
Something occurred to me at dinner... I think the intention here was to _not_ use a secret since the host needs to be on the main IPFS network. Instead, use the new trusted peers mechanism. That seems to work, but so far I haven't been able to get the pin state of the local daemon to reflect in the cluster consensus. More later...

hsanjuan:
> Something occurred to me at dinner... I think the intention here was to _not_ use a secret since the host needs to be on the main IPFS network. Instead, use the new trusted peers mechanism. That seems to work, but so far I haven't been able to get the pin state of the local daemon to reflect in the cluster consensus. More later...

Yes, correct. You will need to set `TrustedPeers` to the peer IDs of your other cafes (suggestions to improve this for your usecase accepted). Cluster secret is not used at all since you already have a host without a network protector.

Can I get write access here, I will have a look to the code and it may be useful now or in the future to commit some changes to the branch (simpler than a PR to the PR :P)

cluster/connector.go
```diff
@@ -0,0 +1,187 @@
...
+		mode, ok := pin.StringToMode(p.Type())
+		if !ok {
+			continue
+		}
+		statusMap[p.Path().String()] = c.pinModeToStatus(mode)
+	}
+	return statusMap, nil
+}
+
+func (c *Connector) ConnectSwarms(ctx context.Context) error {
```
hsanjuan: For information: this will get triggered when a new cafe "bootstraps" (uses `Cluster.Join(<peer maddr>)`). It will be called in the cluster peer it bootstraps to. If textile never calls `Join()` to introduce new peers to the Cluster*, then it can be a noop.

*`Join()` is not strictly necessary with CRDTs, but it is a way of getting two libp2p hosts in the same cluster connected (and this should 1) allow pubsub to start working (otherwise the peer may never receive pubsub messages because it doesn't know any of the other peers subscribed to them), 2) Potentially allow dht-discovery of other cluster peers and better connectivity.

cluster/connector.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return peers, nil
+}
+
+func (c *Connector) ConfigKey(keypath string) (interface{}, error) {
+	return c.node.Repo.GetConfigKey(keypath)
+}
+
+func (c *Connector) RepoStat(ctx context.Context) (*api.IPFSRepoStat, error) {
+	stat, err := corerepo.RepoStat(ctx, c.node)
```
hsanjuan: To check: we call ipfs with `size-only=true`. https://github.com/ipfs/ipfs-cluster/blob/master/ipfsconn/ipfshttp/ipfshttp.go#L600

This makes the RepoStat call WAY faster (no full node count).

cluster/main.go
```diff
@@ -0,0 +1,192 @@
...
+
+	if listenAddr != "" {
+		addr, err := ma.NewMultiaddr(listenAddr)
+		if err != nil {
+			return err
+		}
+		cfgs.ClusterCfg.ListenAddr = addr
+	}
+
+	return cfgMgr.SaveJSON(ConfigPath(repoPath))
```
hsanjuan: I think you don't need to save (it's writing or overwriting to disk). Might be useful to debug what actual config you are generating.

cluster/main.go
```diff
@@ -0,0 +1,192 @@
...
+
+func Bootstrap(ctx context.Context, cluster *ipfscluster.Cluster, cons ipfscluster.Consensus, bootstraps []ma.Multiaddr) {
+	for _, bstrap := range bootstraps {
+		log.Infof("Bootstrapping to %s", bstrap)
+		err := cluster.Join(ctx, bstrap)
+		if err != nil {
+			log.Errorf("bootstrap to %s failed: %s", bstrap, err)
+		} else {
+			for _, p := range cluster.Peers(ctx) {
+				err = cons.Trust(ctx, p.ID)
```
hsanjuan: I guess this is a +1 on auto-trusting peers you bootstrap to. So might get this done in Cluster (it's not yet there).

cluster/main.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+}
+
+func SetupAllocation(
+	name string,
+	diskInfCfg *disk.Config,
+	numpinInfCfg *numpin.Config,
+) (ipfscluster.Informer, ipfscluster.PinAllocator, error) {
+	switch name {
+	case "disk", "disk-freespace":
```
hsanjuan: Probably you want to default to this one. The others are not so useful, but maybe a good placeholder to have this for the future.

cluster/main.go
```diff
@@ -0,0 +1,192 @@
...
+
+func SetupPinTracker(
+	name string,
+	h host.Host,
+	mapCfg *maptracker.Config,
+	statelessCfg *stateless.Config,
+	peerName string,
+) (ipfscluster.PinTracker, error) {
+	switch name {
+	case "map":
```
hsanjuan: for the moment defaulting to this one makes most sense. We want to level up the `stateless` one.

core/cluster.go
```diff
@@ -0,0 +1,114 @@
...
+	informer, alloc, err := cluster.SetupAllocation(
+		"disk-freespace",
+		cfgs.DiskInfCfg,
+		cfgs.NumpinInfCfg,
+	)
+	if err != nil {
+		return err
+	}
+
+	ipfscluster.ReadyTimeout = raft.DefaultWaitForLeaderTimeout + 5*time.Second
```
hsanjuan: I think you can ignore this (like set to 5 seconds or leave default). Raft is not involved anyways.

cmd/main.go
```diff
@@ -436,6 +436,9 @@ Stacks may include:
 	initCafeOpen := initCmd.Flag("cafe-open", "Open the p2p cafe service for other peers").Bool()
 	initCafeURL := initCmd.Flag("cafe-url", "Specify a custom URL of this cafe, e.g., https://mycafe.com").Envar("CAFE_HOST_URL").String()
 	initCafeNeighborURL := initCmd.Flag("cafe-neighbor-url", "Specify the URL of a secondary cafe. Must return cafe info, e.g., via a Gateway: https://my-gateway.yolo.com/cafe, or a cafe API: https://my-cafe.yolo.com").Envar("CAFE_HOST_NEIGHBOR_URL").String()
+	initIpfsCluster := initCmd.Flag("cluster", "Treat the node as an IPFS Cluster peer").Bool()
+	initIpfsClusterBindMultiaddr := initCmd.Flag("cluster-bind-maddr", "Set the IPFS Cluster multiaddrs").Default("/ip4/0.0.0.0/tcp/9096").String()
```
hsanjuan: technically cluster won't bind to anything as it re-uses your already existing ipfs peer.

core/cluster.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	bootstraps, err := cluster.ParseBootstraps(t.config.Cluster.Bootstraps)
+	if err != nil {
+		return err
+	}
+
+	// noop if no bootstraps
+	// if bootstrapping fails, consensus will never be ready
+	// and timeout. So this can happen in background and we
```
hsanjuan: this is not true for CRDT consensus. it will work regardless of bootstrap. Old comment in cluster, we will fix it.

cluster/connector.go
```diff
@@ -0,0 +1,187 @@
...
+		addrs = append(addrs, api.Multiaddr{Multiaddr: addr})
+	}
+	return &api.IPFSID{
+		ID:        c.node.Identity,
+		Addresses: addrs,
+	}, nil
+}
+
+func (c *Connector) SetClient(client *rpc.Client) {
+	// noop
```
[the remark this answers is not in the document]
hsanjuan: Actually, since you don't use rpc at all, it's actually ok like this!

sanderpick:
Thanks for the comments @hsanjuan! You should have write access now.

core/cluster.go
sanderpick: 👌

cmd/main.go
sanderpick: Ah, ok. That makes sense. I can remove this flag then.

core/cluster.go
sanderpick: 👌

cluster/main.go
sanderpick: 👌

cluster/main.go
sanderpick: 👌

cluster/main.go
```diff
@@ -0,0 +1,192 @@
...
+var log = logging.Logger("tex-cluster")
+
+func InitCluster(repoPath, listenAddr string) error {
+	cfgMgr, cfgs := makeClusterConfigs()
+	err := cfgMgr.Default()
+	if err != nil {
+		return err
+	}
+
+	if listenAddr != "" {
```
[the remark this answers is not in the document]
sanderpick: Ok, cool. Should be easy enough to just copy the value used in the IPFS config.

cluster/connector.go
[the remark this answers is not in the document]
sanderpick: 👌

cluster/connector.go
sanderpick: 👌

cluster/main.go
sanderpick: I'll have to double check, but I think this is the first time it's written

cluster/connector.go
```diff
@@ -0,0 +1,187 @@
...
+		})
+		if err != nil {
+			return err
+		}
+		log.Debugf("cluster connected to %s")
+	}
+	return nil
+}
+
+func (c *Connector) SwarmPeers(ctx context.Context) ([]peer.ID, error) {
```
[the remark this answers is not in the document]
sanderpick: Ah, nice that's cool

cluster/main.go
sanderpick: Cool, maybe a config setting to auto-trust?

hsanjuan: Thinking about this. The bootstrap concept in cluster is part of the daemon binary (not of the cluster library itself), so while we can fix this there (https://github.com/ipfs/ipfs-cluster/issues/834), it won't help you.

cluster/main.go
```diff
@@ -0,0 +1,192 @@
...
+	NumpinInfCfg        *numpin.Config
+	TracingCfg          *observations.TracingConfig
+}
+
+func makeClusterConfigs() (*config.Manager, *cfgs) {
+	cfg := config.NewManager()
+	clusterCfg := &ipfscluster.Config{}
+	crdtCfg := &crdt.Config{}
+	maptrackerCfg := &maptracker.Config{}
+	statelessCfg := &stateless.Config{}
```
hsanjuan: I think not worth registering if not using, same for numpinInformer.

PR #850, opened 2019-07-04
Expose a mobile method to fetch CafeRequest by id

asutula:
Really no idea what i'm doing

### DIFF
--- a/mobile/cafes.go
+++ b/mobile/cafes.go
@@ -147,6 +147,24 @@ func (m *Mobile) CafeRequests(limit int) ([]byte, error) {
 	return proto.Marshal(&pb.Strings{Values: groups})
 }
 
+// CafeRequest fetches a single CafeRequest by id
+func (m *Mobile) CafeRequest(id string) ([]byte, error) {
+	if !m.node.Started() {
+		return nil, core.ErrStopped
+	}
+
+	cafeRequest := m.node.Datastore().CafeRequests().Get(id)
+	if cafeRequest == nil {
+		return nil, nil
+	}
+
+	bytes, err := proto.Marshal(cafeRequest)
+	if err != nil {
+		return nil, err
+	}
+	return bytes, nil
+}
+
 // CafeRequestPending marks a request as pending
 func (m *Mobile) CafeRequestPending(id string) error {
 	if !m.node.Started() {
--- a/mobile/cafes_test.go
+++ b/mobile/cafes_test.go
@@ -1,6 +1,7 @@
 package mobile
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -256,9 +257,12 @@ func flushCafeRequests(limit int) (int, error) {
 
 	// make sure we can query for each
 	for g := range reqs {
-		_, err := cafesTestVars.mobile.CafeRequest(g)
+		bytes, err := cafesTestVars.mobile.CafeRequest(g)
 		if err != nil {
 			return count, err
+		}
+		if bytes == nil {
+			return count, errors.New("no bytes returned for CafeRequest")
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


mobile/cafes_test.go
```diff
@@ -254,6 +254,14 @@ func flushCafeRequests(limit int) (int, error) {
 		}
 	}
 
+	// make sure we can query for each
+	for g := range reqs {
+		bytes, err = cafesTestVars.mobile.CafeRequest(g)
```
carsonfarmer: since you haven't created a bytes yet, you'll need to use `bytes, err := cafesTest...`

asutula:
Ok, I think that's a legit failure. It's not returning any bytes when I query for the CafeRequest with my new method. Not sure why, could use any ideas.

asutula:
Could I get some feedback here @sanderpick ?

mobile/cafes.go
```diff
@@ -147,6 +147,24 @@ func (m *Mobile) CafeRequests(limit int) ([]byte, error) {
...
+	if !m.node.Started() {
+		return nil, core.ErrStopped
+	}
+
+	cafeRequest := m.node.Datastore().CafeRequests().Get(id)
+	if cafeRequest == nil {
+		return nil, nil
+	}
+
+	bytes, err := proto.Marshal(cafeRequest)
```
sanderpick: You can just `return proto.Marshal(cafeRequest)`

mobile/cafes_test.go
```diff
@@ -254,6 +255,17 @@ func flushCafeRequests(limit int) (int, error) {
 		}
 	}
 
+	// make sure we can query for each
+	for g := range reqs {
+		bytes, err := cafesTestVars.mobile.CafeRequest(g)
```
sanderpick: Here you're using a group ID where you want a single request ID.

asutula:
In the case of an expired refresh token in the native mobile sdks, we need to ask the host application to re-register the cafe. To do that, we'd need to pass the cafe peer id to the host application. To get the cafe peer id, I need some way of getting a peer id from a cafe request id.

asutula:
Within the context of a failed http request, the only thing I have is the cafe request id. So this would provide a way to map that to cafe information, basically

sanderpick:
Okay, cool! We could add a method like `GetCafeFromGroup` to the datastore interface:

```
type CafeRequestStore interface {
	Queryable
	Add(req *pb.CafeRequest) error
	Get(id string) *pb.CafeRequest
	GetGroup(group string) *pb.CafeRequestList
	GetSyncGroup(group string) string
	Count(status pb.CafeRequest_Status) int
	List(offset string, limit int) *pb.CafeRequestList
	ListGroups(offset string, limit int) []string
	SyncGroupComplete(syncGroupId string) bool
	SyncGroupStatus(groupId string) *pb.CafeSyncGroupStatus
	UpdateStatus(id string, status pb.CafeRequest_Status) error
	UpdateGroupStatus(group string, status pb.CafeRequest_Status) error
	UpdateGroupProgress(group string, transferred int64, total int64) error
	AddAttempt(id string) error
	Delete(id string) error
	DeleteByGroup(groupId string) error
	DeleteBySyncGroup(syncGroupId string) error
	DeleteCompleteSyncGroups() error
	DeleteByCafe(cafeId string) error
}
```

asutula:
Sounds good to me. This was a pretty nieve stab at something like that, so we can close this and do the proper thing.

sanderpick:
Or just make some changes to this?

asutula:
Ok yea sounds good. Will get back on this asap. For now, I'm updating the photos app to just refresh sessions as needed on start up (as a temporary measure)

sanderpick:
Sounds good. If we did that plus changed the token expiration to something like 6 months we could table this for now and focus on trying new features like camera roll sync.

asutula:
Exactly, we're on the same page.

sanderpick:
Awesome. I'm make an issue.

sanderpick:
Still needed @asutula ?

PR #855, opened 2019-07-09
[WIP] Threads2

sanderpick:
So far this just lays out the structure of the new threads model.


carsonfarmer:
This is great @sanderpick, thanks for laying this out. To get the conversation going a bit with potential external collaborators, here's the TL;DR on the access control work here:

Each Thread will have a `Roles` property, which defines a default `Role`, and then peer/account specific roles. The `Role` enum looks like this:

```
enum Role {
    NO_ACCESS = 0;
    READ            = 1;
    ANNOTATE   = 2;
    WRITE          = 3;
}
```

and the `Roles` component looks like this (these are just the protobuf interfaces, but are useful for explanation):

```
message Roles {
    Role default               = 1;
    map<string, Role> accounts = 2;
}
```

This is cool, because you can use `NO_ACCESS` to revoke or prevent access to a specific peer or set of peers. So the default can be `NO_ACCESS`, but with specific peers having (say) `WRITE` access (this would then be a private Thread). Or you can blacklist accounts by specifying specific `NO_ACCESS` roles.

Obviously this stuff is all immutable, so to provide a framework for modifying roles or access, we’ll also be including a simple framework for ‘forking’ Threads. So that way, if you want to (say) add a collaborator with full `WRITE` access, you simply create a new Thread with the right permissions, and point the old Thread to this new Thread id. Since Threads are really just lightweight DAGs, this will be pretty lightweight. Anyone in the old Thread will automatically receive the new Thread keys etc. It also means you can revoke access by specifying a specific peer to have `NO_ACCESS` in the new Thread, and they won’t be able to access the new keys.

Other examples might include say a Thread with a default role of READ, but then perhaps 3 specific addresses with WRITE access. This would be like a public document that is only editable by the ‘owners’. The cool thing here though, is because you’ll be able to fork Threads, anyone with READ access could fork it and add a WRITE role for themselves, but change the default to `NO_ACCESS` (or whatever), thereby creating a new version of the Thread. This makes for some interesting use-cases, like forking a public dataset and adding your own private data…

Another example might be more like a blog, where only the originator gets WRITE access, but the default is `ANNOTATE`. This means, only the originator will be able to write blog posts (top level data items), but anyone will be able to add annotations (items with a target). So you can LIKE or add a MESSAGE to things (we also have a FLAG thread type so the owner can mark comments for removal if they want).

There are lots of other use-cases that we think this new framework will cover, but if you have other ideas, thoughts, concerns, please don't hesitate to leave some comments here or create an issue. We’d like to move pretty quick on this stuff, so any feedback (at least at a conceptual level for the time being) would be useful.

Our goal here is to make sure we’re going to cover enough use-cases to be broadly useful, and with some standard terminology and controls that provide useful abstractions.

carsonfarmer:
Upon further discussion, it looks like we might move schemas and roles to more 'formal' IPLD structures. This would open up a lot of useful properties, but also might have some downsides. It is worth having the discussion about this out in the open here on GitHub.

leshokunin:
Very happy with the suggestions I saw in the email. It's like we discussed a couple months ago. Would work great.
I really dislike the term "annotate" though. Something like "comment" or "react" would sit a lot better for me, but that's just nitpicking.

sanderpick:
I'm not in love w/ "annotate" either (it sounds like machine learning annotations to me), but "commenting" is moving toward an application layer mechanism because the new structure allows for infinite nesting of any block type, i.e., a "comment" is just a message on a message. Similarly, you could "comment" with some data / files, etc. So, an "annotate" privilege seems to capture that behavior most accurately - you're allowed to nest blocks, but not add to the main trunk, so to speak.

leshokunin:
What about React then?

carsonfarmer:
Something else to consider for Threads V2 @sanderpick (from `rvagg ` on #ipld):
> Last comment, if you really are starting a new project, and don't have legacy data concerns, then I'd really really reeeeeeallllly recommend seeing if you can do what needs doing with the new stuff and serialized as dag-cbor.  It's going to have way more first-class support.  And when combined with the plans for IPLD Schemas and representation systems (such as, especially, https://github.com/ipld/specs/blob/165bc97c81a9ef82d3ce1b928e2fea8d6168d7cc/schema-layer/schemas/representations.md#struct-tuple-representation , which lets us make data quite compact if that's a demand that would've otherwise pushed you to protobufs)... well, I just think the story for coming over to CBOR land instead of protobufs should be really compelling unless legacy data is holding you back.

Some additional information there:
> All of the newest work we're doing in IPLD is centered around this concept of the "Data Model" -- https://github.com/ipld/specs/blob/master/data-model-layer/data-model.md -- and then making sure we've specified that satisfactorily, and building all the bridges to various codecs and serial formats and other systems from there.

Its all pretty new, so maybe not worth putting too much effort into right away, but perhaps we should talk with some IPLD core devs are about this soon?

carsonfarmer:
A follow up from tobowers:
> fwiw - protobuf objects in go serialize pretty well to cbor - just one gotcha which is that the standard google/proto stuff has unused fields that end up getting put into cbor structs... but gogo protobuf "faster" and above don't have that and it's possible to strip those fields out in post-processing as well

sanderpick:
@leshokunin doesn't it feel like "REACT" encodes a bit too much of the purpose? As in, what if the intention of nesting is not a reaction, but some other application specific purpose? Maybe something like "BRANCH" could work... "THREAD" would be nice if that wasn't extremely confusing.

sanderpick:
Thanks @carsonfarmer - seems like we should just use cbor nodes throughout unless there are obvious downsides excluding compatibility.

leshokunin:
> 
> 
> @leshokunin doesn't it feel like "REACT" encodes a bit too much of the purpose? As in, what if the intention of nesting is not a reaction, but some other application specific purpose? Maybe something like "BRANCH" could work... "THREAD" would be nice if that wasn't extremely confusing.

I thought about it, but that's really not something that would feel natural. Branch sounds like we're forking the discussion. That's not what's happening here imho.

The reason I like "react" best is because it's somewhat codified in the terminology of Facebook, Slack, Discord: you have the basic object being a message, which you can reply to (comment), but you can also react (put an emoji, "like"). So I think people understand that when there's more than "basic objects" and "commenting", it could be considered "reacting".

I do agree all these terms are super arbitrary and don't encompass everything accurately. I just think "annotate" is basically only used in the context of notes or drawing apps to signify some kind of meta comment.

carsonfarmer:
The only other thing to keep in mind with this naming discussion is that these are backend terms, not application level terms. The application can encode whatever meaning they want/need for the end-user, but it is important to keep the terms as generic as possible for the developer, who might feel 'constrained' if we use application-level terms here. Just my two cents.

PR #856, opened 2019-07-10
Unblock threads

sanderpick:
- Adds a temp solution for unblocking thread listing:
  - Reduce the IPFS timeout to 5 seconds
  - Log an error and return the list of successfully listed threads
- Handles queued blocks concurrently (I noticed that if a post is failing from not being able to cat a block, _all other posts_ will also be blocked)
- Removes the profile publishing that has been happening every time the peer comes online

### DIFF
--- a/core/main.go
+++ b/core/main.go
@@ -377,13 +377,6 @@ func (t *Textile) Start() error {
 		}
 		log.Info("node is online")
 
-		// tmp. publish contact for migrated users.
-		// this normally only happens when peer details are changed,
-		// will be removed at some point in the future.
-		err = t.publishPeer()
-		if err != nil {
-			log.Errorf(err.Error())
-		}
 		go t.cafeOutbox.Flush()
 	}()
 
--- a/core/main.go
+++ b/core/main.go
@@ -555,36 +555,51 @@ func (t *Textile) FlushBlocks() {
 		return util.ProtoTime(pending.Items[i].Date).Before(
 			util.ProtoTime(pending.Items[j].Date))
 	})
-	var posted bool
+	wg := sync.WaitGroup{}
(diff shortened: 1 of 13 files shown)


core/main.go
```diff
@@ -562,36 +555,51 @@ func (t *Textile) FlushBlocks() {
...
-			err := thread.post(block)
-			if err != nil {
-				log.Errorf("error posting block %s: %s", block.Id, err)
-				continue
-			}
-			posted = true
+				err := thread.post(block)
+				if err != nil {
+					log.Errorf("error posting block %s: %s", block.Id, err)
+					if block.Attempts+1 >= maxDownloadAttempts {
```
sanderpick: Adds the attempt handling used when a block is "pending" to when a block is "queued" - this way we can eventually stop trying to post the block.

core/main.go
```diff
@@ -562,36 +555,51 @@ func (t *Textile) FlushBlocks() {
...
-	if posted {
-		go t.cafeOutbox.Flush()
-	} else if t.cafeOutbox.handler != nil {
-		go t.cafeOutbox.handler.Flush()
+				go t.blockOutbox.Flush()
+				if posted {
+					go t.cafeOutbox.Flush()
+				} else if t.cafeOutbox.handler != nil {
+					go t.cafeOutbox.handler.Flush()
+				}
```
sanderpick: These lines of code that deal with "flushing" are nasty - we need a way to flush a single block instead of queuing up a bunch of flushes that attempt to empty the entire queue - it's just heavy handed

sanderpick:
The go.mod changes are just from running build - not sure if IPFS has tweaked their dependencies or if this just never got run before the last release.

PR #858, opened 2019-07-11
Include failing thread in list

sanderpick:
One more tweak here @asutula - I realized this temp solution is pretty useless if you're just left in a state where it always takes 5 seconds to list your threads and you're unable to delete the bad thread. So, this just includes the failed thread but w/o it's `head_blocks`, which are not loading from the local repo. 

In the UI, you could check for at least one `head_blocks` item. If missing, render an "alert" icon or something that indicated to the user that they should not expect normal behavior from this thread - in practice, this could just lead you to delete it.


asutula:
Nice @sanderpick. This is actually in line with an idea I was going to write up today in an issue: Basically that when we're dealing with data that is loaded in an indeterministic way, if possible, we should return some placeholder with error state object. As you said, this is good because the data the user expects to see isn't just mysteriously absent.

PR #877, opened 2019-07-26
Last (hopefully) thread v1 improvements

sanderpick:
- Includes block ciphertext in the thread envelope in an attempt to ensure it can be available to recipients
- Adds back direct messaging when possible, but retains the pubsub ack mechanism as step 2 in a 3-step delivery process:
  1. If swarm-connected, push the message over the existing connection 
  2. If not, publish the message and wait for an ack (this may leverage a relay node)
  3. If offline or no ack, add to outbox for cafe delivery


core/thread.go
```diff
@@ -713,7 +713,7 @@ func (t *Thread) sendWelcome() error {
 		if err != nil {
 			return err
 		}
-		env, err := t.service().NewEnvelope(t.Id, ndata, sig)
+		env, err := t.service().NewEnvelope(t.Id, ndata, ciphertext, sig)
```
carsonfarmer: Nice

Issue #879, opened 2019-07-29
Remove 'default' global flags from daemon, init, and wallet commands


carsonfarmer:
Right now, if you run help for say `textile daemon`, it shows several 'global' flags that don't actually apply to the daemon command:
```
Flags:
      --help              Show context-sensitive help (also try --help-long and --help-man).
      --api="http://127.0.0.1:40600"  
                          API Address to use
      --api-version="v0"  API version to use
      --debug             Set the logging level to debug
```

These should be 'turned off' for the `daemon`, `init`, and `wallet` subcommands, which don't actually use these flags. Otherwise, their presence is a bit confusing...

PR #890, opened 2019-08-06
globals_flags: fix proposal #879

ilpaijin:
### 879 fix proposal
It seems it needs to be a hook PreAction which disables the unwanted flags declared in the main Application cmd.
Before parsing, choose an arbitrary list of cmds (it could have been a single call per cmd, or others...I didn't get you way of design it so I simply choose one way)

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -903,6 +903,12 @@ The response contains a base58 encoded version of the random bytes token.`).Alia
 
 	// ================================
 
+	hideGlobalsFlagsFor(
+		daemonCmd,
+		initCmd,
+		walletCmd,
+	)
+
 	// commands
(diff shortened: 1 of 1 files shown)


sanderpick:
Hey @ilpaijin - This looks like a reasonable approach, thanks! Can you add a sign-off to your commit? 

```
git commit --amend --signoff
git push --force-with-lease origin issue/879
```

ilpaijin:
sure @sanderpick, done. If you don't mind I'll try to check for other issues, some easy to pick for me.

sanderpick:
Oh, it would be nice to allow `--help` to still come through for `init`, `daemon`, and `wallet`. Maybe you can filter that out of the hide step.

As for working on other issues, of course... any help is much appreciated.

ilpaijin:
ok nice tx

PR #892, opened 2019-08-12
remove some problematic and unneeded locks

asutula:


andrewxhill:
I'm still seeing some blocks in Dev at least.

Two clear places.

1. Join a Cafe with some photos already in your wallet. In this case, the UI will block until all photos have finished syncing. Ouch for more than a couple photos.
2. Send a message (text is fine). The keyboard will lock until the message is posted in the thread. Makes a boggy feeling keyboard for message->message->message rattling.

Can I fix this with something in the app / rn sdk?

sanderpick:
For the first one, making thread joining async would solve it... not sure about the second one. Maybe adding a message is not async either? Will have to look.

sanderpick:
Ok, I took a look. I'm remembering now that we started the async push with just adding files and data. Here are the remaining methods that could be changed if they're causing issues (I'll put a star next to the ones that we should do ASAP):

- `AddMessage`*
- `AddLike`*
- `AddComment`*
- `AddIgnore`*
- `AddFlag`

- `AcceptInvite`*
- `AcceptExternalInvite`*
- `IgnoreInvite`*
- `AddInvite`
- `AddExternalInvite`

- `SetName`*

- `AddSchema`

- `AddContact`
- `RemoveContact`

- `AcceptInviteViaNotification`*
- `IgnoreInviteViaNotification`*

- `AddThread`*
- `AddOrUpdateThread`*
- `RenameThread`*
- `RemoveThread`*
- `SnapshotThreads`*

PR #900, opened 2019-09-23
Scope repo path with account seed

asutula:
This is step 1 for supporting account switching/log out

### DIFF
--- a/core/cafes_test.go
+++ b/core/cafes_test.go
@@ -15,36 +15,36 @@ import (
 )
 
 var cafeVars = struct {
-	nodePath string
-	cafePath string
+	nodeInitConfig InitConfig
+	cafeInitConfig InitConfig
 
 	node *Textile
 	cafe *Textile
 
 	token string
 }{
-	nodePath: "./testdata/.textile3",
-	cafePath: "./testdata/.textile4",
+	nodeInitConfig: InitConfig{
+		BaseRepoPath: "./testdata/.textile3",
+		Debug:        true,
+	},
+	cafeInitConfig: InitConfig{
+		BaseRepoPath: "./testdata/.textile4",
+		Debug:        true,
+		SwarmPorts:   "4001",
+		CafeApiAddr:  "0.0.0.0:5000",
+		CafeURL:      "http://127.0.0.1:5000",
+		CafeOpen:     true,
+	},
 }
 
 func TestCore_SetupCafes(t *testing.T) {
 	var err error
-	cafeVars.node, err = CreateAndStartPeer(InitConfig{
-		RepoPath: cafeVars.nodePath,
-		Debug:    true,
-	}, true)
+	cafeVars.node, err = CreateAndStartPeer(cafeVars.nodeInitConfig, true)
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	cafeVars.cafe, err = CreateAndStartPeer(InitConfig{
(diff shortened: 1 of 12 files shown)


core/core.go
```diff
@@ -119,9 +119,19 @@ var ErrStarted = fmt.Errorf("node is started")
...
+// RepoPath returns the actual location of the configured repo
+func (conf InitConfig) RepoPath() string {
+	return path.Join(conf.BaseRepoPath, conf.Account.Seed())
+}
+
+// RepoExists return whether or not the configured repo already exists
+func (conf InitConfig) RepoExists() bool {
+	return fsrepo.IsInitialized(conf.RepoPath())
+}
+
```
asutula: These methods for `InitConfig` provide a way to get the full repo path (base path + seed) and check if the repo is initialized already. Equivalent methods exist in the mobile module.

core/test_utils.go
```diff
@@ -14,15 +14,16 @@ import (
 )
 
 func CreateAndStartPeer(conf InitConfig, wait bool) (*Textile, error) {
-	_ = os.RemoveAll(conf.RepoPath)
-
 	conf.Account = keypair.Random()
+
+	_ = os.RemoveAll(conf.RepoPath())
```
asutula: repo path is derived from the account, so we need to move this until after the account is created

core/core.go
```diff
@@ -53,7 +53,7 @@ const kSyncAccountFreq = time.Hour
 type InitConfig struct {
 	Account         *keypair.Full
 	PinCode         string
-	RepoPath        string
+	BaseRepoPath    string
```
asutula: Now called `BaseRepoPath` to denote that it's not the actual repo path, but used as a base location. Actual location is determined by account seed.

mobile/mobile.go
```diff
@@ -92,6 +95,16 @@ type Mobile struct {
...
+// RepoPath returns the actual location of the configured repo
+func (conf InitConfig) RepoPath() string {
+	return path.Join(conf.BaseRepoPath, conf.Seed)
+}
+
+// RepoExists return whether or not the configured repo already exists
+func (conf InitConfig) RepoExists() bool {
+	return fsrepo.IsInitialized(conf.RepoPath())
+}
+
```
asutula: Similar updates to the mobile module. I considered proxying through to the core version of `InitConfig` so that `path` and `fsrepo` packages wouldn't have to be imported in the mobile module, but it makes the mobile API nastier because these methods would have to return `(<value>, error)` instead of a single value. Not worth it.

repo/init.go
```diff
@@ -135,7 +135,7 @@ func checkWriteable(dir string) error {
 
 	if os.IsNotExist(err) {
 		// dir doesnt exist, check that we can create it
-		return os.Mkdir(dir, 0775)
+		return os.MkdirAll(dir, 0775)
```
asutula: This was necessary in order to create the parent dir which is the account seed.

core/cafes_test.go
```diff
@@ -15,36 +15,36 @@ import (
...
+	cafeInitConfig InitConfig
 
 	node *Textile
 	cafe *Textile
 
 	token string
 }{
-	nodePath: "./testdata/.textile3",
-	cafePath: "./testdata/.textile4",
+	nodeInitConfig: InitConfig{
```
sanderpick: Nice!

PR #903, opened 2019-09-25
Scope repo with account address instead of seed

asutula:


mobile/mobile.go
```diff
@@ -96,13 +93,50 @@ type Mobile struct {
 }
 
 // RepoPath returns the actual location of the configured repo
-func (conf InitConfig) RepoPath() string {
-	return path.Join(conf.BaseRepoPath, conf.Seed)
+func (conf InitConfig) RepoPath() (string, error) {
```
asutula: `RepoPath()` and `RepoExists()` are a little less nice to use if we scope the repo path with account address instead of seed. This is because we have to convert the provided seed to an account, which can fail, so we need to bubble that error up.

mobile/mobile.go
```diff
@@ -96,13 +93,50 @@ type Mobile struct {
...
-	return fsrepo.IsInitialized(conf.RepoPath())
+func (conf InitConfig) RepoExists() (bool, error) {
+	coreConf, err := conf.coreInitConfig()
+	if err != nil {
+		return false, err
+	}
+	return coreConf.RepoExists(), nil
+}
+
+func (conf InitConfig) account() (*keypair.Full, error) {
```
asutula: Helper to get an account object from a mobile `InitConfig`

mobile/mobile.go
```diff
@@ -96,13 +93,50 @@ type Mobile struct {
...
+		return nil, err
+	}
+	accnt, ok := kp.(*keypair.Full)
+	if !ok {
+		return nil, keypair.ErrInvalidKey
+	}
+	return accnt, nil
+}
+
+func (conf InitConfig) coreInitConfig() (core.InitConfig, error) {
```
asutula: Helper to get a `core.InitConfig` from a `mobile.InitConfig`

PR #904, opened 2019-09-26
Use base repo and account seed in cli

asutula:
Gets the CLI in line with how the programatic API deals with repo locations:

- Removes `repo` flag from `init` and uses `base-repo` instead. It's optional and defaults to `~/.textile/repo`.
- The `daemon` flag now uses `base-repo` instead of `repo` with the same default.
- The `daemon` flag takes an optional `account-seed` flag to specify which account to use. If left out, it defaults to the first sub directory inside `base-repo`
- Same thing for `migrate` flag

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -1224,6 +1224,9 @@ func getRepo(baseRepo string, accountSeed string) (string, error) {
 		if len(files) == 0 {
 			return "", fmt.Errorf("no account repos initialized in: %s", baseRepo)
 		}
+		if len(files) > 1 {
+			return "", fmt.Errorf("there are multiple accounts initialzed in %s, you should specify account-seed", baseRepo)
+		}
 		return path.Join(baseRepo, files[0].Name()), nil
 	}
 	kp, err := keypair.Parse(accountSeed)
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -331,7 +331,7 @@ An access token is required to register, and should be obtained separately from 
 	// daemon
 	daemonCmd := appCmd.Command("daemon", "Start a node daemon session")
 	daemonBaseRepo := daemonCmd.Flag("base-repo", "Specify a custom path to the base repo directory").Short('b').String()
-	daemonAccountSeed := daemonCmd.Flag("account-seed", "Specify an existing account seed").Short('a').String()
+	daemonAccountAddress := daemonCmd.Flag("account-address", "Specify an existing account address").Short('a').String()
 	daemonPin := daemonCmd.Flag("pin", "Specify the pin code for datastore encryption (omit no pin code was used during init)").Short('p').String()
 	daemonDocs := daemonCmd.Flag("serve-docs", "Whether to serve the local REST API docs").Short('s').Bool()
 	cmds[daemonCmd.FullCommand()] = func() error {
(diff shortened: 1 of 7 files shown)


cmd/cmd.go
```diff
@@ -329,11 +330,16 @@ An access token is required to register, and should be obtained separately from
 
 	// daemon
 	daemonCmd := appCmd.Command("daemon", "Start a node daemon session")
-	daemonRepo := daemonCmd.Flag("repo", "Specify a custom repository path").Short('r').String()
+	daemonBaseRepo := daemonCmd.Flag("base-repo", "Specify a custom path to the base repo directory").Short('b').String()
+	daemonAccountSeed := daemonCmd.Flag("account-seed", "Specify an existing acconut seed").Short('a').String()
```
sanderpick: can this be address? it sounds insecure to have to paste the private key into the terminal on daemon start

sanderpick: (typo in "acconut")

cmd/cmd.go
```diff
@@ -1197,9 +1208,33 @@ func getRepo(repo string) (string, error) {
...
+			return "", err
+		}
+		if len(files) == 0 {
+			return "", fmt.Errorf("no account repos initialized in: %s", baseRepo)
+		}
+		return path.Join(baseRepo, files[0].Name()), nil
+	}
+	kp, err := keypair.Parse(accountSeed)
+	if err != nil {
+		return "", fmt.Errorf(fmt.Sprintf("parse account seed failed: %s", err))
```
sanderpick: nit, Errorf takes a format string, so the inner Sprintf is not required

asutula:
Added a little optimization where, if you there are multiple account initialized in the base repo location and you don't specify an account seed, it fails and tells you to provide an account seed. So, the only time you can omit account seed is when there is only one account initialized.

cmd/cmd.go
asutula: Possibly. Let me look. My initial thinking was you have to be able to derive an actual `Account` (which I think you can only do from seed), but maybe an `Account` isn't required when starting the daemon.

asutula: Either way, remember you have to provide you account seed into the terminal in order to initialize a repo

sanderpick: Right, but that's a one time deal, after which the seed is stored (and encrypted if you want) in the repo. The seed isn't needed to start the daemon/node, here you're just using it to choose the repo path.

asutula: Yea makes sense. This also applies to migrate. I'll update it now.

sanderpick: sweet

cmd/cmd.go
asutula: Cool, fixed it here and everywhere else in the repo

sanderpick: oh nice, i must have been the origin of that ha

cmd/cmd.go
asutula: done.

asutula:
Ok @sanderpick re: lock file... what do you think the lock file will be used for?

asutula:
To update the original PR description with the current state of affairs:

- The daemon flag takes an optional `account-address` flag to specify which account to use. If left out, it defaults to the single sub directory inside base-repo, or errors out if there is more than one
- Same thing for migrate flag

sanderpick:
A file (not really a lock file)... would be handy under the base repo to determine which repo is "active" / logged in to. This way you wouldn't have to always specify account address when starting the node, which requires that state to be kept somewhere (in my 1pass or in the containing app, for example). You could just write the account address into this file.

asutula:
Ok I understand the idea. Would that state be updated any time you _do_ specify an account address to `daemon` and/or `migrate`? 

It's not really blocking this current work, so maybe I'll add this now, but maybe save it for another PR?

sanderpick:
yeah, specifying the address would mean "change account". agreed, not blocking this one...

PR #905, opened 2019-09-26
bots: adds the first implementation of bots rpc

andrewxhill:
Signed-off-by: andrew <andrew@textile.io>


core/bot_service.go
```diff
@@ -0,0 +1,236 @@
...
+
+// BotKVStore is the kv store interface shared with a bot. Set, Get, Delete required
+type BotKVStore struct {
+	botID      string
+	botVersion int
+	node       *Textile
+}
+
+// Get allows a bot to get IPFS data by the cid/path. Allows optional key for decryption on the fly
+func (mip BotIpfsHandler) Get(pth string, key string) ([]byte, error) {
```
sanderpick: why not a pointer method?

core/bot_service.go
```diff
@@ -0,0 +1,236 @@
...
+	"path"
+
+	"github.com/mr-tron/base58/base58"
+	shared "github.com/textileio/go-textile-core/bots"
+	"github.com/textileio/go-textile/crypto"
+	ipfs "github.com/textileio/go-textile/ipfs"
+)
+
+// BotIpfsHandler is the ipfs interface shared with a bot. It requires Get and Add
+type BotIpfsHandler struct {
```
sanderpick: does this implement and interface in go-textile-core?

core/core.go
```diff
@@ -278,6 +280,11 @@ func NewTextile(conf RunConfig) (*Textile, error) {
 	return node, nil
 }
 
+// Start creates an ipfs node and starts textile services
```
sanderpick: doc string isn't right

core/core.go
```diff
@@ -366,6 +373,14 @@ func (t *Textile) Start() error {
 	if err != nil {
 		return err
 	}
+
+	// get ready to run some bots
+	log.Debug("creating the bot service...")
+	t.botService = NewBotService(
+		t,
+	)
```
sanderpick: `NewBotService(t)` ?

repo/datastore.go
```diff
@@ -63,6 +64,13 @@ type PeerStore interface {
 	DeleteByAddress(address string) error
 }
 
+type BotStore interface {
```
sanderpick: This could be moved in to the core interfaces along with this whole `Datastore` interface (eventually).

repo/migrations/migration016.go
```diff
@@ -0,0 +1,58 @@
...
+		return err
+	}
+	if pinCode != "" {
+		if _, err := db.Exec("pragma key='" + pinCode + "';"); err != nil {
+			return err
+		}
+	}
+
+	query := `
+		drop table botstore;
```
sanderpick: formatting nit

core/bot_service.go
andrewxhill: it does yeah, https://github.com/textileio/go-textile-core/blob/master/bots/interface.go#L29, i've updated my doc string in upcoming push

core/bot_service.go
andrewxhill: well... short answer, i don't know. worth exploring more but i banged my head against the wall for a bit when first figuring that out. the plugin does not like it when i use pointers for these shared methods

repo/datastore.go
andrewxhill: 👍

andrewxhill:
okay, got each of the comments.

re. moving client / server to bots. sounds good. i can give you BotClient => botsrepo.Client easy (done here), Server is trickier. Currently it is relying on the *core IPFS node, so if I move it out and then depend on it, it creates a circular dep.

sanderpick:
K, cool. Makes sense. Eventually it would be nice to pull out the pieces you need from `Textile` (IPFS, datastore), and then make sure they are referenced by core interfaces (`Datastore`, `IPFS`). Then you'd build the Server from those. Then, it could all be assembled in Textile. For another day.

PR #911, opened 2019-10-04
migration006 fix, base repo fix

andrewxhill:


andrewxhill:
cc @asutula, when you come back you should take a look at this and be sure that the baserepo check wont cause anything strange on mobile

asutula:
Took a look @andrewxhill. So that change around supporting pre-0.7.3 repos only effects the cli (and derived cli things like the docker image), shouldn't effect mobile at all.

But... we need to deal with the same thing in mobile because we have existing installs with the old style repos. I'll start working through that.

PR #916, opened 2019-10-07
start of cmd for add/remove/list bots

andrewxhill:
- exposes external methods of bots over cafe api. this fixes some of the demo issues I had re. mapping internal apis via nginx. instead, i just pass the cafe requests to the bot requests for non-internal methods
- moves bots api route slightly to make the cafe api capable of passing non-internal requests more easily
- adds the first cmd method `list`.


api/api.go
```diff
@@ -329,10 +329,13 @@ func (a *Api) Run() {
 
 		bots := v0.Group("/bots")
 		{
-			bots.GET("/:root", a.botsGet)
-			bots.POST("/:root", a.botsPost)
-			bots.DELETE("/:root", a.botsDelete)
-			bots.PUT("/:root", a.botsPut)
+			bots.GET("/list", a.botsList)
+			bots.GET("/disable", a.botsDisable)
+			bots.GET("/enable", a.botsEnable)
```
sanderpick: these look like POST requests

core/cafe_api_v1.go
```diff
@@ -632,6 +634,23 @@ func (c *cafeApi) search(g *gin.Context) {
 	})
 }
 
+func ReverseProxyBotApi(method string) gin.HandlerFunc {
+	return func(g *gin.Context) {
+		id := g.Param("id")
+		s := fmt.Sprintf("/api/v0/bots/id/%s", id)
+		u, _ := url.Parse("localhost:40600") // TODO get dynamic port
```
sanderpick: if this method was part of Cafe, you could snag the port there

andrewxhill:
just updated to go-textile-core to get those latest models, will rebase master now too

PR #917, opened 2019-10-08
Support the old style repo config

asutula:
Along side new account-scoped repo directories, we can let the docker files and library users continue to specify `repo` if they chose. This will buy us some time to get the account scoped repos api correct and complete.

The only change is there is no default behavior if you don't specify repo path OR base repo path and account... you have to supply one or the other. This is to avoid likely errors that will in happen in practice if there is a default behavior.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -638,7 +638,7 @@ There are two types of invites, direct account-to-account and external:
 
 	// migrate
 	migrateCmd := appCmd.Command("migrate", "Migrate the node repository and exit")
-	migrateRepo := migrateCmd.Flag("repo", "Specify a custom path to the repo directory").Short('b').String()
+	migrateRepo := migrateCmd.Flag("repo", "Specify a custom path to the repo directory").Short('r').String()
 	migrateBaseRepo := migrateCmd.Flag("base-repo", "Specify a custom path to the base repo directory").Short('b').String()
 	migrateAccountAddress := migrateCmd.Flag("account-address", "Specify an existing account address").Short('a').String()
 	cmds[migrateCmd.FullCommand()] = func() error {
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -14,14 +14,12 @@ import (
 	"os"
 	"os/signal"
 	"path"
-	"path/filepath"
 	"strings"
 
 	"github.com/fatih/color"
 	"github.com/golang/protobuf/jsonpb"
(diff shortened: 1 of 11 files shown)


cmd/cmd.go
```diff
@@ -646,14 +638,11 @@ There are two types of invites, direct account-to-account and external:
 
 	// migrate
 	migrateCmd := appCmd.Command("migrate", "Migrate the node repository and exit")
+	migrateRepo := migrateCmd.Flag("repo", "Specify a custom path to the repo directory").Short('b').String()
```
sanderpick: short version => `r` ?

cmd/cmd.go
```diff
@@ -1193,49 +1182,20 @@ func output(val interface{}) {
...
-			for _, file := range files {
-				if !file.IsDir() && file.Name() == "textile" {
-					// This is a pre 0.7.2 repo, not subdir under account path
-					return baseRepo, nil
-				}
-			}
-			return "", fmt.Errorf("there are multiple accounts initialzed in %s, you need to specify account-address", baseRepo)
-		}
-		return path.Join(baseRepo, files[0].Name()), nil
+	if err := os.MkdirAll(finalRepo, 0755); err != nil {
```
sanderpick: In the case where you have an existing repo folder and are not using a base repo, this may result in the same docker issue we saw with doodle. It would be nice to check this behavior.

asutula: ok, I'll try to test it out.

asutula: I ended up just removing all the code in `cmd.go` that attempts to create directories... it's not needed because all the directory creation is handled elsewhere in the initialization process. I'll still make sure this works with the Dockerfiles

asutula: Ok, seems to run fine in docker. I ran these commands to test:
```
docker build --file=./Dockerfile --tag=textile .
docker run textile
docker build --file=./Dockerfile.cafe --tag=textile-cafe .
docker run textile-cafe
```

asutula:
Ok @sanderpick this could be g2g. Give it a look.

asutula:
Oh, another thing... For the cli, I ended up providing a default behavior if you don't specify a repo or base repo/account. In that case, for `initialize`, `migrate` and `daemon` it defaults to the `~/.textile/repo` location. I chose to go that way because so many docs already assume that behavior and it will continue to work the way most existing users are used to.

PR #923, opened 2019-10-09
build: adds env var to build CGO_ENABLED=0 

andrewxhill:
ensures that the build of the CI machine is also statically linked


sanderpick:
Wait, we need CGO for SQLite.

andrewxhill:
ahh.. there's the rub

PR #943, opened 2019-11-28
Use same go + node version on CircleCI

maxnordlund:
Something I noticed when reading through the source.

This probably needs more work, but it's a start I hope :)


maxnordlund:
No problem, always happy to help :)

Den tis 3 dec. 2019 17:28Sander Pick <notifications@github.com> skrev:

